PR #14, opened 2023-03-24
Add VFS support

Godones:
Try to introduce support for vfs, which is referenced from linux, and currently supports some common functions. I was not sure how this architecture should use the filesystem, so I wrote only a few simple functions and did not use disk devices, because the vfs implementation includes a ramfs that can run in memory.

If this request is merged, I will add a supported version of vfs for fat32, which will use block devices.

Please tell me how this system needs to support the file system?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .DS_Store
 *.asm
 *.img
+.idea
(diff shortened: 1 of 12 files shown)


equation314:
Thank you very much!

We have introduced the filesystem in https://github.com/rcore-os/arceos/pull/19, your contribution has inspired us a lot.

We are sorry that this PR will be closed.

PR #19, opened 2023-04-09
Add filesystem framework

equation314:
As title.


modules/axfs/src/fs/fatfs.rs
```diff
@@ -0,0 +1,262 @@
...
+        match pos {
+            SeekFrom::Start(pos) => {
+                self.set_position(pos);
+                Ok(pos)
+            }
+            SeekFrom::Current(off) => {
+                let pos = (self.position() as i64 + off) as u64;
+                self.set_position(pos);
+                Ok(pos)
+            }
```
yfblock, line 236: Such as SeekFrom::Current(position: isize), Is it possible that it is less than 0?

yfblock (approved):

yfblock (approved):

PR #52, opened 2023-05-24
Add riscv hypervisor support

KuangjuX:
This PR accomplished the following:

1. Referenced hypercraft and added hypervisor support for the riscv target.
2. Modified axruntime to support hypervisor.
3. Modified libax to add hypervisor app interfaces.
4. Created hv app to apply guest.

However, there are some issues to be resolved:

1. x86 and aarch64 are still not supported and currently rely on dummy to pass the compilation.
2. Currently using an old version of rust toolchain, as seen in the issue: https://github.com/rust-lang/rust/issues/111637


chyyuu:
please fix the missing doc comments

PR #56, opened 2023-05-27
Add Ext2 filesystem to arceos

deathwings602:
This PR do the following:
+ Add a crate `ext2fs`, which implements Ext2 filesystem.
+ Add a crate `ext2fs_fuse`, used for create Ext2 filesystem image.
+ Modify `axfs` so that it can use `ext2fs` feature to enable Ext2 filesystem as main filesystem.
+ Modify `apps/fs/shell` and `crates/axfs_vfs` to support symbolic link, link and recursive deletion.


chyyuu:
@deathwings602 please fix the doc error

PR #88, opened 2023-06-09
Loadbalance pr3

131131yhx:
New feature(s):

1. Split the run_queue into one queue per hart.
2. Add loadbalance crate (with some modifications in scheduler crate) with a Zircon style implementation.
3. Support automatically choice of hart to run for each new tasks.
4. Support thread steal mechanism between different harts.
5. Add new task called unbalance. It has many unbalance payloads to run, simulating real world parallel apps like SpMV.
Achievement(s):
6. The thread steal mechanism speed up by 10% in task unbalance. Moreover, current version is slightly faster than single run_queue even with only 4 cores in unbalance task.
7. The high concurrence of I/O thread wake up simulation obtains more than 50% time reduction on SMP=8 ACCEL=y compared to single run queue


equation314:
Is this PR still active? Please fix conflicts.

PR #90, opened 2023-06-09
Unitest

chunchunyd:
App_tests on qemu-aarch64 all time out but I can't find the reasons.


equation314:
Why are there 8,000,000+ modifications?

chunchunyd:
@equation314 
The reason for so many modifications:
1. Because the kernel  cannot access the network in online evaluation environment of the preliminary, we used the cargo vendor command to download all third-party dependencies, more than 200 of them. Most of the added codes came from here. We will modify the .gitignore file later; 
2. The dev-monolithic branch currently leads the main branch by more than 300 commits, and it is very difficult for our project to merge the work of other groups' 300+ commits into this branch in a short time. Because we do not have permission to directly modify the branch, and considering that this branch should not be merged into dev in a short time like dev-microkernel, in order to eliminate conflicts, I directly overwrote dev-monolithic with main. Most of the deleted codes came from here. Later we will gradually merge the codes of other students, but can we temporarily change dev-monolithic to checkout from the main branch?

PR #105, opened 2023-07-23
Refactor ulib

equation314:
- move cbindings out of `libax` as a new crate `axlibc`
- rename `libax` to `axstd`
- make APIs in `axstd` more similar to rust std, now apps can be built with both std and ArceOS without modification
- update the architecture diagram

TODO:
- [ ]  axfeat
- [ ]  arceos_api
- [ ] arceos_posix_api


chyyuu (changes requested):
suggest :   scope of  arceos_api > scope of arceos_api_posix , so     "arceos_api"  --> "arceos_std_api" ;  musllic is used for linux, not only posix,  "arceos_api_posix" --> "arceos_linux_api"

equation314:
I think the linux API is at the syscall level, and the POSIX API is at the library function level. The ArceOS API layer is not only syscalls, but also includes high-level library functions, and it is more appropriately called `arceos_api_posix`.

Called `arceos_api` instead of `arceos_std_api` because the name is too long.

equation314:
`arceos_api` -> `arceos_rust_api` ?

chyyuu (approved):

PR #115, opened 2023-08-23
alloc: update memory allocation pointer type from NonZeroUsize to NonNull<u8> in ByteAllocator trait

pymongo:
This pull request updates the memory allocation pointer type from NonZeroUsize to NonNull<u8> in the ByteAllocator trait of the Rust open-source operating system project.

Motivation/Improvement Reason:

1. The NonZeroUsize type can represent memory page numbers, allocation sizes, or offsets, which can be ambiguous in terms of semantics. Using NonNull<u8> provides a more accurate representation of returning a pointer to a memory segment.
2. Previously, both the tlsf and buddy libraries required conversion from NonZeroUsize to NonNull<u8>. After this change, only the slab library needs to perform the conversion.
3. The Rust standard library's core::alloc::Alloc also uses NonNull<u8>.

This PR modifies the ByteAllocator trait and its implementations to use NonNull<u8> as the memory allocation pointer type. The changes include updating the return type of the alloc method to AllocResult<NonNull<u8>>, and the parameter type of the dealloc method to NonNull<u8>

### DIFF
--- a/api/arceos_api/src/imp/mem.rs
+++ b/api/arceos_api/src/imp/mem.rs
@@ -4,13 +4,13 @@ cfg_alloc! {
 
     pub fn ax_alloc(layout: Layout) -> Option<NonNull<u8>> {
         if let Ok(vaddr) = axalloc::global_allocator().alloc(layout) {
-            Some(unsafe { NonNull::new_unchecked(vaddr.get() as _) })
+            Some(vaddr)
         } else {
             None
         }
     }
 
     pub fn ax_dealloc(ptr: NonNull<u8>, layout: Layout) {
-        axalloc::global_allocator().dealloc(ptr.addr(), layout)
+        axalloc::global_allocator().dealloc(ptr, layout)
     }
 }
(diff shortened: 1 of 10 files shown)


equation314:
Thanks! Please fix CI.

equation314:
After using `NonNull<u8>` instead of `NonZeroUsize`, the crate attributes `#![feature(strict_provenance)]` can be removed. Could you find all the attributes and remove them?

api/arceos_api/src/imp/mem.rs
```diff
@@ -4,13 +4,13 @@ cfg_alloc! {
 
     pub fn ax_alloc(layout: Layout) -> Option<NonNull<u8>> {
         if let Ok(vaddr) = axalloc::global_allocator().alloc(layout) {
-            Some(unsafe { NonNull::new_unchecked(vaddr.get() as _) })
+            Some(vaddr)
```
equation314: These lines can be simplified to `axalloc::global_allocator().alloc(layout).ok()`

equation314 (approved):

PR #121, opened 2023-09-07
feat: add new command ldr and str to load/store to memory

limingth:
add 2 command ldr & str in apps/fs/shell for debugging raspi4 register

### DIFF
--- a/apps/cli/src/cmd.rs
+++ b/apps/cli/src/cmd.rs
@@ -1,0 +1,155 @@
+use std::io::{self};
+
+#[cfg(all(not(feature = "axstd"), unix))]
+
+macro_rules! print_err {
+    ($cmd: literal, $msg: expr) => {
+        println!("{}: {}", $cmd, $msg);
+    };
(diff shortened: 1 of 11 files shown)


equation314:
Memory dumping should not be the responsibility of a `shell` application, it would be better to put such functions in the `tools` directory.

apps/cli/src/cmd.rs
```diff
@@ -0,0 +1,155 @@
...
+    println!("ldr");
+    if args.is_empty() {
+        println!("try: ldr ffff0000400fe000 / ldr ffff000040080000 ffff000040080008");
+    }
+
+    fn ldr_one(addr: &str) -> io::Result<()> {
+        println!("addr = {}", addr);
+
+        if let Ok(parsed_addr) = u64::from_str_radix(addr, 16) {
+            let address: *const u64 = parsed_addr as *const u64; // 强制转换为合适的指针类型
```
hky1999, line 63: Annotation by Chinese is not a good idea.

PR #159, opened 2024-07-23
Separate apps into the arceos-apps repository

equation314:
https://github.com/arceos-org/arceos-apps

# TODO

* Simplify build & run scripts.
* Make apps download the scripts only instead of the entire ArceOS repository, avoiding the use of `make -C path/to/arceos`.


modules/axdriver/src/drivers.rs
```diff
@@ -1,6 +1,6 @@
 //! Defines types and probe methods of all supported devices.
 
-#![allow(unused_imports)]
+#![allow(unused_imports, dead_code)]
```
hky1999, line 3: Why allow dead_code here

equation314: Eliminate warnings on `make clippy`.

PR #163, opened 2024-07-29
Introduce module axmm for virtual memory management

equation314:


modules/axmm/src/aspace.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /// Add a new linear mapping.
+    ///
+    /// The mapping is linear, i.e., `start_vaddr` is mapped to `start_paddr`,
+    /// and `start_vaddr + size` is mapped to `start_paddr + size`.
+    ///
+    /// The `flags` parameter indicates the mapping permissions and attributes.
+    ///
+    /// Returns an error if the address range is out of the address space or not
```
hky1999, line 62: delete "or not"

modules/axmm/src/aspace.rs
```diff
@@ -0,0 +1,139 @@
...
+                false, // flush_tlb_by_page
+            )
+            .map_err(paging_err_to_ax_err)?
+            .flush();
+        Ok(())
+    }
+
+    /// Removes mappings within the specified virtual address range.
+    ///
+    /// Returns an error if the address range is out of the address space or not
```
hky1999, line 95: delete "or not"
Returns an error if the address range is out of the address space or with bad alignment.

modules/axmm/src/aspace.rs
hky1999: Returns an error if the address range is out of the address space or with bad alignment.

hky1999 (approved):

PR #164, opened 2024-07-30
Refactor trap handler with linkme::distributed_slice

equation314:
It allows trap handlers to be defined in any crate.


modules/axhal/src/trap.rs
```diff
@@ -1,23 +1,31 @@
...
-///
-/// [1]: crate_interface::def_interface
-/// [2]: crate_interface::impl_interface
-#[def_interface]
-pub trait TrapHandler {
-    /// Handles interrupt requests for the given IRQ number.
-    fn handle_irq(irq_num: usize);
-    // more e.g.: handle_page_fault();
-}
+pub use linkme::distributed_slice as reg_trap_handler;
```
hky1999: what does 'reg' in `reg_trap_handler` mean?

scripts/make/cargo.mk
```diff
@@ -17,7 +17,7 @@ build_args := \
   $(build_args-$(MODE)) \
   $(verbose)
 
-RUSTFLAGS := -C link-arg=-T$(LD_SCRIPT) -C link-arg=-no-pie
+RUSTFLAGS := -C link-arg=-T$(LD_SCRIPT) -C link-arg=-no-pie -C link-arg=-znostart-stop-gc
```
hky1999, line 20: I this place need an annotation, like add refs about what is 'nostart-stop-gc', I find something here: https://releases.llvm.org/14.0.0/tools/lld/docs/ELF/start-stop-gc.html

ulib/axstd/src/os.rs
```diff
@@ -3,4 +3,6 @@
 /// ArceOS-specific definitions.
 pub mod arceos {
     pub use arceos_api as api;
+    #[doc(no_inline)]
+    pub use arceos_api::modules;
```
hky1999, line 7: Previously, ning did not expose these modules. What was ning's consideration for exposing all of them in the API now? Is there any safety concern if methods or vars in those modules is not exported as `pub(crate)`

modules/axhal/src/irq.rs
```diff
@@ -33,3 +34,11 @@ pub(crate) fn register_handler_common(irq_num: usize, handler: IrqHandler) -> bo
     warn!("register handler for IRQ {} failed", irq_num);
     false
 }
+
+#[reg_trap_handler(IRQ)]
```
hky1999: no `#[reg_trap_handler(PAGE_FAULT)]` currently ,right?

hky1999 (changes requested):

modules/axhal/src/trap.rs
equation314: register

scripts/make/cargo.mk
equation314: https://github.com/dtolnay/linkme/issues/94

ulib/axstd/src/os.rs
equation314: To use `axhal` in apps without adding it to dependencies.

modules/axhal/src/irq.rs
equation314: Not required for the unikernels.

hky1999 (approved):

PR #166, opened 2024-08-02
Update dependencies versions

equation314:


modules/axhal/Cargo.toml
```diff
@@ -55,7 +55,7 @@ aarch64-cpu = "9.4"
 tock-registers = "0.8"
 arm_gicv2 = "0.1"
 arm_pl011 = "0.1"
-arm_pl031 = { version = "0.1", optional = true }
+arm_pl031 = { version = "0.2", optional = true }
```
hky1999, line 58: as well as percpu crate

percpu = "0.1" => percpu = "0.1.3"

hky1999 (changes requested):

modules/axhal/Cargo.toml
equation314: It already pinned in `Cargo.lock`

hky1999: get it.
lgtm

hky1999 (approved):

PR #167, opened 2024-08-04
[mm]: Implement protect

tkf2019:


modules/axmm/src/aspace.rs
```diff
@@ -160,11 +160,22 @@ impl AddrSpace {
             return ax_err!(InvalidInput, "address not aligned");
         }
 
-        // TODO
-        self.pt
-            .protect_region(start, size, flags, true)
-            .map_err(|_| AxError::BadState)?
-            .ignore();
+        self.areas
+            .protect(
```
equation314, line 164: How to handle `UNCACHE` flag changes:  https://github.com/arceos-org/arceos/blob/dcf728208bc47dcc610e3d181317b46d9170b196/modules/axdma/src/dma.rs#L59

PR #170, opened 2024-09-07
feat: support percpu run-queue and cpu affinity for axtask

hky1999:


hky1999:
TO BE FIXED: 

`make A=../arceos-apps/c/pthread/pipe SMP=4 LOG=info ARCH=aarch64 ACCEL=n run`

This will cause a  deadlock.

hky1999:
@equation314 @aarkegz Need help, there is a bug in https://github.com/arceos-org/arceos/actions/runs/10794370098/job/29938498068?pr=170,

However, it is so so so difficult to reproduce.

I think it is related with `exit_current`, because all tasks have completed the computing and exited, but main task failed to get back running after it `join` on all other worker tasks.

hky1999:
scheduler https://github.com/arceos-org/scheduler/pull/1

PR #173, opened 2024-09-14
Introduce user-defined task extended data

equation314:
Major changes to `axtask`:

1. Add a `task_ext` field to `TaskInner`, and allow users to define the specific structure via the `def_task_ext!` macro.
2. Add a `spawn_task` API with `TaskInner` as the argument.

### DIFF
--- a/modules/axruntime/src/mp.rs
+++ b/modules/axruntime/src/mp.rs
@@ -1,6 +1,7 @@
-use axconfig::{SMP, TASK_STACK_SIZE};
-use axhal::mem::{va, virt_to_phys};
 use core::sync::atomic::{AtomicUsize, Ordering};
+
+use axconfig::{SMP, TASK_STACK_SIZE};
+use axhal::mem::{virt_to_phys, VirtAddr};
 
 #[link_section = ".bss.stack"]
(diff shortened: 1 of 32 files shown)


modules/axruntime/src/mp.rs
```diff
@@ -11,7 +12,7 @@ pub fn start_secondary_cpus(primary_cpu_id: usize) {
     let mut logic_cpu_id = 0;
     for i in 0..SMP {
         if i != primary_cpu_id {
-            let stack_top = virt_to_phys(va!(unsafe {
+            let stack_top = virt_to_phys(VirtAddr::from(unsafe {
```
hky1999, line 15: you just revert it back (dog head manually

hky1999 (approved):

modules/axruntime/src/mp.rs
aarkegz: 🙃

aarkegz (approved):

modules/axruntime/src/mp.rs
equation314: I removed `va` and `pa` exports in `axhal::mem`, so we need to import the `memory_addr` dependency if use `va!`.

Do you think it is OK?

hky1999: I agree with the former

PR #176, opened 2024-09-20
[task] support percpu run-queue and cpu affinity for axtask

hky1999:
scheduler : https://github.com/arceos-org/scheduler/pull/1


.github/workflows/build.yml
```diff
@@ -33,6 +33,7 @@ jobs:
       continue-on-error: ${{ matrix.rust-toolchain == 'nightly' }}
       run: make clippy ARCH=aarch64
     - name: Check code format
+      continue-on-error: ${{ matrix.rust-toolchain == 'nightly' }}
```
equation314: Update to the latest `main` branch.

modules/axtask/src/run_queue.rs
```diff
@@ -1,72 +1,249 @@
...
+///
+/// The index (cpu_id) of the selected run queue.
+///
+/// ## Panics
+///
+/// This function will panic if there is no available run queue that matches the CPU set.
+///
+#[cfg(feature = "smp")]
+#[inline]
+fn select_run_queue_index(cpumask: CpuMask<{ axconfig::SMP }>) -> usize {
```
equation314: Add a type alias:

```rs
type CpuMask = cpumask::CpuMask<{ axconfig::SMP }>;
```

modules/axtask/src/run_queue.rs
```diff
@@ -1,72 +1,249 @@
...
+/// A reference to the selected `AxRunQueue`.
+///
+/// ## TODO
+///
+/// 1. Implement better load balancing across CPUs for more efficient task distribution.
+/// 2. Use a more generic load balancing algorithm that can be customized or replaced.
+///
+#[inline]
+pub(crate) fn select_run_queue<G: BaseGuard>(
+    #[cfg(feature = "smp")] task: AxTaskRef,
```
equation314: Please avoid using conditional compilation in function arguments.

modules/axtask/src/run_queue.rs
```diff
@@ -1,72 +1,249 @@
...
         let curr = crate::current();
-        if !curr.is_idle() && self.scheduler.task_tick(curr.as_task_ref()) {
+        if !curr.is_idle() && self.inner.scheduler.task_tick(curr.as_task_ref()) {
             #[cfg(feature = "preempt")]
             curr.set_preempt_pending(true);
         }
     }
 
     pub fn yield_current(&mut self) {
+        let _kernel_guard = kernel_guard::NoPreemptIrqSave::new();
```
equation314: Duplicate guard?

modules/axtask/src/run_queue.rs
```diff
@@ -1,72 +1,249 @@
...
+        G::release(self.state);
+    }
+}
+
 impl AxRunQueue {
-    pub fn new() -> SpinNoIrq<Self> {
+    pub fn new(cpu_id: usize) -> Self {
         let gc_task = TaskInner::new(gc_entry, "gc".into(), axconfig::TASK_STACK_SIZE).into_arc();
+        // gc task shoule be pinned to the current CPU.
+        gc_task.set_cpumask(CpuMask::from_usize(1 << cpu_id));
```
equation314: The `from_usize` API is bad, change to `from_raw_bits(1 << cpu_id)` or `oneshot(cpu_id)`.

modules/axtask/src/run_queue.rs
equation314: Can we get rid of the brackets around `axconfig::SMP`?

modules/axtask/src/run_queue.rs
hky1999: ok, it'll be fixed soon

modules/axtask/src/run_queue.rs
hky1999: ok

modules/axtask/src/run_queue.rs
hky1999: ao, forget to delete it.

modules/axtask/src/run_queue.rs
hky1999:
```bash
error[E0573]: expected type, found constant `axconfig::SMP`
  --> /home/hky/workspace/arceos-org/arceos/modules/axtask/src/run_queue.rs:87:44
   |
87 | fn select_run_queue_index(cpumask: CpuMask<axconfig::SMP>) -> usize {
   |                                            ^^^^^^^^^^^^^ not a type

error[E0747]: unresolved item provided when a constant was expected
  --> /home/hky/workspace/arceos-org/arceos/modules/axtask/src/run_queue.rs:87:44
   |
87 | fn select_run_queue_index(cpumask: CpuMask<axconfig::SMP>) -> usize {
   |                                            ^^^^^^^^^^^^^
   |
help: if this generic argument was intended as a const parameter, surround it with braces
   |
87 | fn select_run_queue_index(cpumask: CpuMask<{ axconfig::SMP }>) -> usize {
   |                                            +               +

Some errors have detailed explanations: E0573, E0747.
For more information about an error, try `rustc --explain E0573`.
error: could not compile `axtask` (lib) due to 2 previous errors
make: *** [scripts/make/build.mk:34: _cargo_build] Error 101
```

hky1999: https://users.rust-lang.org/t/why-do-const-generics-need-braces-when-the-path-of-a-constant-is-composed-of/111240

Cargo.lock
```diff
@@ -1022,9 +1031,9 @@ checksum = "2dffe52ecf27772e601905b7522cb4ef790d2cc203488bbd0e2fe85fcb74566d"
 
 [[package]]
 name = "memory_addr"
-version = "0.3.1"
+version = "0.3.0"
```
equation314: Get rid of this change.

modules/axtask/src/timers.rs
```diff
@@ -1,40 +1,57 @@
...
-static TIMER_LIST: LazyInit<SpinNoIrq<TimerList<TaskWakeupEvent>>> = LazyInit::new();
+use crate::{select_run_queue, AxTaskRef};
 
-struct TaskWakeupEvent(AxTaskRef);
+percpu_static! {
+    TIMER_LIST: LazyInit<TimerList<TaskWakeupEvent>> = LazyInit::new(),
+}
+
+struct TaskWakeupEvent {
+    ticket_id: usize,
```
equation314: Use `u64` here, it can easily overflow.

modules/axtask/src/task.rs
```diff
@@ -39,9 +46,28 @@ pub struct TaskInner {
...
+
+    /// Mark whether the task is in the wait queue.
     in_wait_queue: AtomicBool,
+    /// Mark whether the task is in the timer list.
     #[cfg(feature = "irq")]
     in_timer_list: AtomicBool,
+    /// A ticket ID used to identify the timer event.
+    /// Incremented by 1 each time the timer event is triggered or expired.
+    #[cfg(feature = "irq")]
+    timer_ticket_id: AtomicUsize,
```
equation314: Merge `in_timer_list` and `timer_ticket_id`? `timer_ticket_id == 0` indicates it is not in the timer list.

modules/axtask/src/task.rs
```diff
@@ -199,6 +233,7 @@ impl TaskInner {
     pub(crate) fn new_init(name: String) -> Self {
         let mut t = Self::new_common(TaskId::new(), name);
         t.is_init = true;
+        t.set_cpumask(CpuMask::one_shot(this_cpu_id()));
```
equation314: Why not use `CpuMask::full()`?

modules/axtask/src/api.rs
```diff
@@ -14,6 +16,9 @@ pub use crate::wait_queue::WaitQueue;
 /// The reference type of a task.
 pub type AxTaskRef = Arc<AxTask>;
 
+/// The wrapper type for cpumask::CpuMask with SMP configuration.
+pub type CpuSet = cpumask::CpuMask<{ axconfig::SMP }>;
```
equation314: Using two different names can be misleading. Why not: 

```rs
pub type CpuMask = cpumask::CpuMask<{ axconfig::SMP }>;
```

modules/axtask/src/task.rs
```diff
@@ -266,6 +310,50 @@ impl TaskInner {
...
+        debug!("{} unblocking", self.id_name());
+
+        // If the owning (remote) CPU is still in the middle of schedule() with
+        // this task as prev, wait until it's done referencing the task.
+        //
+        // Pairs with the `clear_prev_task_on_cpu()`.
+        //
+        // This ensures that tasks getting woken will be fully ordered against
+        // their previous state and preserve Program Order.
+        if self.on_cpu() {
```
equation314: `if` is unnecessary?

modules/axtask/src/run_queue.rs
```diff
@@ -1,45 +1,216 @@
...
+/// This function returns a static reference to the current run queue, which
+/// is inherently unsafe. It assumes that the `RUN_QUEUE` has been properly
+/// initialized and is not accessed concurrently in a way that could cause
+/// data races or undefined behavior.
+///
+/// ## Returns
+///
+/// A static reference to the current run queue.
+// #[inline(always)]
+pub(crate) fn current_run_queue<G: BaseGuard>() -> AxRunQueueRef<'static, G> {
```
equation314, line 58: Use default generic type? 

```rs
fn current_run_queue<G: BaseGuard = NoOp>()
```

Cargo.lock
hky1999: ok

modules/axtask/src/timers.rs
hky1999: ok, but what is the difference between usize and u64 under 64bit ?

modules/axtask/src/task.rs
hky1999: No, we don't think we can simply use `timer_ticket_id == 0` to indicate a task is not in time list.

`timer_ticket_id` means current availble timer id, it has to be incremented each time a timer is triggered or expired.

If we set `timer_ticket_id` to 0, we cannot determine whether the previous timer event for this task was triggered by the timer or by a notification.

* The timer ticket ID is used to address this issue: the original timer event has been canceled, but before the deadline is reached, a new timer event occurs, setting in_timer_list to true. When the deadline for the original timer event arrives, it will be triggered.
*  The reason I keep the `in_timer_list` flag is to determine in the wait queue's `cancel_events`, whether to call `timer_ticket_expire_one`, instead of simply incrementing the timer ticket ID every time in `cancel_events`.

hky1999: see `modules/axtask/src/wait_queue.rs`
```Rust
    fn cancel_events(&self, curr: CurrentTask) {
        // A task can be wake up only one events (timer or `notify()`), remove
        // the event from another queue.
        if curr.in_wait_queue() {
            // wake up by timer (timeout).
            let mut wq_locked = self.queue.lock();
            wq_locked.retain(|t| !curr.ptr_eq(t));
            curr.set_in_wait_queue(false);
        }
        #[cfg(feature = "irq")]
        if curr.in_timer_list() {
            curr.set_in_timer_list(false);
            curr.timer_ticket_expire_one();
            // TODO:
            // this task is still not removed from timer list of target CPU,
            // which may cause some redundant timer events.
        }
    }
```

modules/axtask/src/task.rs
hky1999: I think **init task** should be only scheduled on this core, the `exit_current` will judge if it's the init task that exits, to determine if the system should be shutdown.
```Rust
 pub fn exit_current(&mut self, exit_code: i32) -> ! {
        let curr = crate::current();
        debug!("task exit: {}, exit_code={}", curr.id_name(), exit_code);
        assert!(curr.is_running(), "task is not running: {:?}", curr.state());
        assert!(!curr.is_idle());
        if curr.is_init() {
            EXITED_TASKS.with_current(|exited_tasks| exited_tasks.clear());
            axhal::misc::terminate();
        } else {
        }
```

modules/axtask/src/api.rs
hky1999: yep, y r right

modules/axtask/src/task.rs
hky1999: it refers to implementations in skyloft
```C
    /* task must be scheduled atomically */
    if (unlikely(atomic_load_acq(&next->stack_busy))) {
        /* wait until the scheduler finishes switching stacks */
        while (atomic_load_acq(&next->stack_busy)) cpu_relax();
    }
```

And I think it can be cooperated with the assertion in the end of this `if` code block

modules/axtask/src/run_queue.rs
hky1999: Tried, but:
```bash
error: defaults for type parameters are only allowed in `struct`, `enum`, `type`, or `trait` definitions
  --> /home/hky/workspace/arceos-org/arceos/modules/axtask/src/run_queue.rs:58:33
   |
58 | pub(crate) fn current_run_queue<G: BaseGuard = NoOp>() -> AxRunQueueRef<'static, G> {
   |                                 ^^^^^^^^^^^^^^^^^^^
   |
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #36887 <https://github.com/rust-lang/rust/issues/36887>
   = note: `#[deny(invalid_type_param_default)]` on by default
```

modules/axtask/src/timers.rs
```diff
@@ -1,40 +1,57 @@
...
 
 impl TimerEvent for TaskWakeupEvent {
     fn callback(self, _now: TimeValue) {
-        let mut rq = RUN_QUEUE.lock();
-        self.0.set_in_timer_list(false);
-        rq.unblock_task(self.0, true);
+        // Ignore the timer event if the task is not in the timer list.
+        // timeout was set but not triggered (wake up by `WaitQueue::notify()`).
+        // Judge if this timer event is still valid by checking the ticket ID.
+        if !self.task.in_timer_list() || self.task.timer_ticket() != self.ticket_id {
```
Azure-stars: Is the `!self.task.in_timer_list()` necessary? When a timer event that has been marked expired (`cancel_events` is executed) arrives, we should ignore this event according to semantics. At this time, its ticket_id should be different from the ticket_id of the current task (because `timer_ticket_expire_one` is executed). It seems that there is no need to use the in_timer_list attribute.

modules/axtask/src/run_queue.rs
```diff
@@ -84,49 +257,51 @@ impl AxRunQueue {
...
 
-        curr.set_state(TaskState::Blocked);
-        wait_queue_push(curr.clone());
-        self.resched(false);
+        debug!("task block: {}", curr.id_name());
+        self.inner.resched(false);
     }
 
+    /// Unblock one task by inserting it into the run queue.
+    /// If task state is `BLOCKING`, it will enter a loop until the task is in `BLOCKED` state.
```
Azure-stars: Fix the comment for `BLOCKING`

modules/axtask/src/run_queue.rs
```diff
@@ -84,49 +257,51 @@ impl AxRunQueue {
...
     pub fn unblock_task(&mut self, task: AxTaskRef, resched: bool) {
-        debug!("task unblock: {}", task.id_name());
-        if task.is_blocked() {
+        task.clone().unblock_locked(|| {
+            let cpu_id = self.inner.cpu_id;
+            debug!("task unblock: {} on run_queue {}", task.id_name(), cpu_id);
             task.set_state(TaskState::Ready);
-            self.scheduler.add_task(task); // TODO: priority
-            if resched {
+            self.inner.scheduler.lock().add_task(task.clone()); // TODO: priority
```
Azure-stars: Can we use `put_prev_task` instead of `add_task`? Because `add_task` will generate a new `vruntime` for the task, and it may not be conducive to time slice statistics

modules/axtask/src/timers.rs
hky1999: Yep, you are right, I've thought about this before.

The thing is, I do not want to waste a timer ticket ID due to a NOT-EXISTED timer event from `cancel_events`, which is called each time when a task is exiting `wait_xxx` function, even through that task is not put into timer_list. That why I reseved the `if curr.in_timer_list()` judgement in `cancel_events`.

hky1999: Maybe we can divide `cancel_events` into two parts? @equation314

modules/axtask/src/run_queue.rs
hky1999: Thx, it will be fixed soon

modules/axtask/src/run_queue.rs
hky1999: It depends on the semantic of `vruntime` for the task in CFS scheduling

modules/axtask/src/timers.rs
equation314: For 32-bit platforms in future.

modules/axtask/src/task.rs
```diff
@@ -254,16 +292,61 @@ impl TaskInner {
...
+    }
+
+    pub(crate) fn unblock_locked<F>(&self, mut run_queue_push: F)
+    where
+        F: FnMut(),
+    {
+        debug!("{} unblocking", self.id_name());
+
+        // When irq is enabled, use `unblock_lock` to protect the task from being unblocked by timer and `notify()` at the same time.
+        #[cfg(feature = "irq")]
```
equation314: Remove this? How about two processors unblocks the same task at the same time?

modules/axtask/src/run_queue.rs
```diff
@@ -1,45 +1,215 @@
...
+fn select_run_queue_index(cpumask: CpuMask) -> usize {
+    static RUN_QUEUE_INDEX: AtomicUsize = AtomicUsize::new(0);
+
+    assert!(!cpumask.is_empty(), "No available CPU for task execution");
 
-// TODO: per-CPU
-static EXITED_TASKS: SpinNoIrq<VecDeque<AxTaskRef>> = SpinNoIrq::new(VecDeque::new());
+    // Round-robin selection of the run queue index.
+    loop {
+        let index = RUN_QUEUE_INDEX.load(Ordering::SeqCst) % axconfig::SMP;
```
equation314: Fetch and add here?

modules/axtask/src/run_queue.rs
```diff
@@ -1,45 +1,215 @@
...
+///
+/// A reference to the `AxRunQueue` corresponding to the provided index.
+///
+/// ## Panics
+///
+/// This function will panic if the index is out of bounds.
+///
+#[inline]
+fn get_run_queue(index: usize) -> &'static mut AxRunQueue {
+    assert!(index < axconfig::SMP);
```
equation314: `assert` is duplicated, The out-of-bounds check is performed at array indexing.

modules/axtask/src/run_queue.rs
```diff
@@ -1,45 +1,215 @@
...
+///
+/// ## Panics
+///
+/// This function will panic if `cpu_mask` is empty, indicating that there are no available CPUs for task execution.
+///
+#[cfg(feature = "smp")]
+// The modulo operation is safe here because `axconfig::SMP` is always greater than 1 with "smp" enabled.
+#[allow(clippy::modulo_one)]
+#[inline]
+fn select_run_queue_index(cpumask: CpuMask) -> usize {
```
equation314, line 88: Use reference to `cpumask`

modules/axtask/src/run_queue.rs
```diff
@@ -1,45 +1,215 @@
...
+        let index = select_run_queue_index(task.cpumask());
+        AxRunQueueRef {
+            inner: get_run_queue(index),
+            state: irq_state,
+            _phantom: core::marker::PhantomData,
+        }
+    }
+}
 
+/// AxRunQueue represents a run queue for global system or a specific CPU.
```
equation314: Add ` `` `

modules/axtask/src/run_queue.rs
```diff
@@ -1,45 +1,215 @@
...
+    fn drop(&mut self) {
+        G::release(self.state);
+    }
 }
 
 impl AxRunQueue {
-    pub fn new() -> SpinNoIrq<Self> {
+    pub fn new(cpu_id: usize) -> Self {
         let gc_task = TaskInner::new(gc_entry, "gc".into(), axconfig::TASK_STACK_SIZE).into_arc();
+        // gc task shoule be pinned to the current CPU.
```
equation314: `should`

modules/axtask/src/run_queue.rs
```diff
@@ -1,45 +1,215 @@
...
+
+impl<'a, G: BaseGuard> Drop for AxRunQueueRef<'a, G> {
+    fn drop(&mut self) {
+        G::release(self.state);
+    }
 }
 
 impl AxRunQueue {
-    pub fn new() -> SpinNoIrq<Self> {
+    pub fn new(cpu_id: usize) -> Self {
```
equation314: Merge the two implementation blocks of `AxRunQueue`

modules/axtask/src/run_queue.rs
```diff
@@ -84,49 +257,56 @@ impl AxRunQueue {
...
+            self.inner.resched(false);
         }
         unreachable!("task exited!");
     }
 
-    pub fn block_current<F>(&mut self, wait_queue_push: F)
-    where
-        F: FnOnce(AxTaskRef),
-    {
+    pub fn blocked_resched(&mut self) {
```
equation314: Why call it `blocked_resched`?

modules/axtask/src/task.rs
```diff
@@ -244,6 +273,15 @@ impl TaskInner {
         self.is_idle
     }
 
+    #[inline]
+    pub(crate) fn cpumask(&self) -> CpuMask {
```
equation314, line 290: Return reference for better performance.

modules/axtask/src/wait_queue.rs
```diff
@@ -81,23 +106,39 @@ impl WaitQueue {
...
+            debug!("{} push to wait queue on wait_until", curr.id_name());
+
+            // we must not block current task with preemption disabled.
+            // Current expected preempt count is 2.
+            // 1 for `NoPreemptIrqSave`, 1 for wait queue's `SpinNoIrq`.
+            #[cfg(feature = "preempt")]
+            assert!(curr.can_preempt(2));
+            wq.push_back(curr.clone());
+
+            curr.set_state(TaskState::Blocked);
```
equation314: Try to put all task state manipulation to `run_queue.rs`

modules/axtask/src/run_queue.rs
```diff
@@ -84,49 +257,56 @@ impl AxRunQueue {
...
     }
 
+    /// Unblock one task by inserting it into the run queue.
+    /// If task's `on_cpu` flag is true,
+    /// it will enter a loop until the task finishes its scheduling process,
+    /// see `unblock_locked` for details.
     pub fn unblock_task(&mut self, task: AxTaskRef, resched: bool) {
-        debug!("task unblock: {}", task.id_name());
-        if task.is_blocked() {
+        task.clone().unblock_locked(|| {
```
equation314: Can we put `fn unblock_locked` here?

modules/axtask/src/run_queue.rs
```diff
@@ -153,25 +333,41 @@ impl AxRunQueue {
...
+                prev_task.id_name(),
+                next_task.id_name()
+            );
+        }
         #[cfg(feature = "preempt")]
         next_task.set_preempt_pending(false);
         next_task.set_state(TaskState::Running);
+        // Claim the task as running, we do this before switching to it
+        // such that any running task will have this set.
+        next_task.set_on_cpu(true);
```
equation314: Move it after `prev_task.ptr_eq(next_task)`

modules/axtask/src/task.rs
hky1999: BUT it's wrapped by a `SpinNoIrq`

modules/axtask/src/task.rs
hky1999: A task CAN NOT be unblocked by two processors at the same time, right?

Is there any situation where a task exists in two wait queues concurrently?

modules/axtask/src/run_queue.rs
hky1999: Corresponding to `preempt_resched`, it is primarily called by the `wait_xxx` methods in `wait_queue.rs` to check whether the current task is in a `BLOCKED` state.

modules/axtask/src/wait_queue.rs
hky1999: Trying, we have to make sure task's state is set as `Blocked` with holding the lock of wait queue!

modules/axtask/src/run_queue.rs
hky1999: make sense, done

modules/axtask/src/task.rs
hky1999: detailed comments have been added

modules/axtask/src/run_queue.rs
hky1999: Yep, but `unblock_lock` is a field of axtask structure, I do not want to expose it to other mods

modules/axtask/src/run_queue.rs
```diff
@@ -1,45 +1,198 @@
...
+///
+/// A reference to the selected `AxRunQueue`.
+///
+/// ## TODO
+///
+/// 1. Implement better load balancing across CPUs for more efficient task distribution.
+/// 2. Use a more generic load balancing algorithm that can be customized or replaced.
+///
+#[inline]
+pub(crate) fn select_run_queue<G: BaseGuard>(task: AxTaskRef) -> AxRunQueueRef<'static, G> {
```
equation314: `task: AxTaskRef` -> `task: &AxTaskRef`

api/axfeat/Cargo.toml
```diff
@@ -13,7 +13,7 @@ documentation = "https://arceos-org.github.io/arceos/axfeat/index.html"
 default = []
 
 # Multicore
-smp = ["axhal/smp", "axruntime/smp", "kspin/smp"]
```
equation314, line 16: `axtask/smp` -> `axtask?/smp`

api/axfeat/Cargo.toml
```diff
@@ -13,7 +13,7 @@ documentation = "https://arceos-org.github.io/arceos/axfeat/index.html"
...
+smp = ["axhal/smp", "axruntime/smp", "axtask/smp", "kspin/smp"]
```
equation314: `"kspin/smp"` can be removed, it has been added in `axtask/Cargo.toml`

modules/axtask/src/task.rs
```diff
@@ -39,9 +45,25 @@ pub struct TaskInner {
     entry: Option<*mut dyn FnOnce()>,
     state: AtomicU8,
 
+    /// CPU affinity mask.
+    cpumask: SpinNoIrq<CpuMask>,
+
+    /// Mark whether the task is in the wait queue.
     in_wait_queue: AtomicBool,
+    /// A ticket ID used to identify the timer event.
+    /// Incremented by 1 each time the timer event is triggered or expired.
```
equation314: Wrong comment

modules/axtask/src/task.rs
```diff
@@ -39,9 +45,25 @@ pub struct TaskInner {
...
+    #[cfg(feature = "irq")]
+    timer_ticket_id: AtomicU64,
```
equation314: Put it near `unblock_lock`

modules/axtask/src/task.rs
```diff
@@ -320,6 +410,50 @@ impl TaskInner {
...
+            .store(false, Ordering::Release);
+    }
+
+    /// Returns whether the task is running on a CPU.
+    ///
+    /// It is used to protect the task from being moved to a different run queue
+    /// while it has not finished its scheduling process.
+    /// The `on_cpu field is set to `true` when the task is preparing to run on a CPU,
+    /// and it is set to `false` when the task has finished its scheduling process in `clear_prev_task_on_cpu()`.
+    pub fn on_cpu(&self) -> bool {
```
equation314: Put it near `set_on_cpu`

modules/axtask/src/task.rs
```diff
@@ -1,19 +1,20 @@
 use alloc::{boxed::Box, string::String, sync::Arc};
 use core::ops::Deref;
-use core::sync::atomic::{AtomicBool, AtomicI32, AtomicU64, AtomicU8, Ordering};
+#[cfg(any(feature = "preempt", feature = "irq"))]
+use core::sync::atomic::AtomicUsize;
```
equation314: Not used?

modules/axtask/src/task.rs
```diff
@@ -429,8 +563,11 @@ impl Deref for CurrentTask {
 }
 
 extern "C" fn task_entry() -> ! {
-    // release the lock that was implicitly held across the reschedule
-    unsafe { crate::RUN_QUEUE.force_unlock() };
+    unsafe {
+        // Clear the prev task on CPU before running the task entry function.
+        crate::current().clear_prev_task_on_cpu();
```
equation314, line 552: Why clear it? It will not used before we call `set_prev_task_on_cpu` on context switch.

hky1999: A switch to B, and task B is not running before, task B should begin with task_entry (not the last line of switch_to),task B is responsible for clear the on_cpu flag of task A, right?

modules/axtask/src/wait_queue.rs
```diff
@@ -27,49 +29,54 @@ use crate::{AxRunQueue, AxTaskRef, CurrentTask, RUN_QUEUE};
...
+        // Try to cancel a timer event from timer lists.
+        // Just mark task's current timer ticket ID as expired.
         #[cfg(feature = "irq")]
-        if curr.in_timer_list() {
-            // timeout was set but not triggered (wake up by `WaitQueue::notify()`)
-            crate::timers::cancel_alarm(curr.as_task_ref());
+        if from_timer_list {
+            curr.timer_ticket_expired();
+            // TODO:
+            // this task is still not removed from timer list of target CPU,
```
equation314: This comment is outdated.

modules/axtask/src/run_queue.rs
```diff
@@ -140,101 +352,161 @@ impl AxRunQueue {
...
+    idle_task.set_cpumask(CpuMask::one_shot(cpu_id));
     IDLE_TASK.with_current(|i| {
         i.init_once(idle_task.into_arc());
     });
 
     // Put the subsequent execution into the `main` task.
     let main_task = TaskInner::new_init("main".into()).into_arc();
     main_task.set_state(TaskState::Running);
-    unsafe { CurrentTask::init_current(main_task) };
+    main_task.set_on_cpu(true);
```
equation314: `main` task is an init task, we should set full cpumask.

modules/axtask/src/task.rs
equation314: Each task except the init task will start after `switch_to`

modules/axtask/src/wait_queue.rs
```diff
@@ -81,23 +88,23 @@ impl WaitQueue {
     where
         F: Fn() -> bool,
     {
+        let mut rq = current_run_queue::<NoPreemptIrqSave>();
```
equation314: We should not disable IRQ around the loop.

modules/axtask/src/wait_queue.rs
```diff
@@ -156,9 +163,11 @@ impl WaitQueue {
     /// If `resched` is true, the current task will be preempted when the
     /// preemption is enabled.
     pub fn notify_one(&self, resched: bool) -> bool {
-        let mut rq = RUN_QUEUE.lock();
-        if !self.queue.lock().is_empty() {
-            self.notify_one_locked(resched, &mut rq)
+        let mut wq = self.queue.lock();
+        if let Some(task) = wq.pop_front() {
+            task.set_in_wait_queue(false);
```
equation314: Put this line in `unblock_one_task`

api/axfeat/Cargo.toml
hky1999: Sorry, it's a mistake

modules/axtask/src/wait_queue.rs
```diff
@@ -186,31 +195,28 @@ impl WaitQueue {
...
-            true
-        } else {
-            false
-        }
-    }
-
-    pub(crate) fn notify_one_locked(&self, resched: bool, rq: &mut AxRunQueue) -> bool {
-        if let Some(task) = self.queue.lock().pop_front() {
+        let task_to_be_notify = {
+            if let Some(index) = wq.iter().position(|t| Arc::ptr_eq(t, task)) {
```
equation314: Merge these two `if`s.

modules/axtask/src/task.rs
hky1999: Really?
A task only continutes its running after `switch_to` when it has been running before,
otherwise it should start with `task_entry`, am I wrong?

modules/axtask/src/wait_queue.rs
hky1999: I DO NOT think so, we just mark this timer event as expired, instead of removing it from timer list, it will still trigger a timer event, enter `callback` and be ignored due to Timer ticket ID not matched.

modules/axtask/src/task.rs
```diff
@@ -244,6 +280,15 @@ impl TaskInner {
         self.is_idle
     }
 
+    #[inline]
+    pub(crate) fn cpumask(&self) -> CpuMask {
+        *self.cpumask.lock()
+    }
+
+    pub(crate) fn set_cpumask(&self, cpumask: CpuMask) {
```
equation314, line 295: Missing `#[inline]`

modules/axtask/src/run_queue.rs
hky1999: As I said here https://github.com/arceos-org/arceos/pull/176#discussion_r1780458568 , init task should be pinned on this CPU

modules/axtask/src/task.rs
equation314: `task_entry` is written to `ctx.rip` and the task will jump here after `swtich_to`.

modules/axtask/src/wait_queue.rs
equation314: It can be considered a lazy-removal strategy, ignoring it when it is about to take effect.

api/axfeat/Cargo.toml
equation314: Sorry, we cannot remove `kspin/smp`, consider the `smp` feature is on and the `multitask` feature is off, we should enable `smp` in `kspin`.

modules/axtask/src/run_queue.rs
equation314: Idle tasks should be pinned rather than init tasks.

hky1999: okay, I understand that

modules/axtask/src/wait_queue.rs
hky1999: However, it still hasn't been completely removed and needs to go through the process of expiring an event from the timer list and invoking the callback.

modules/axtask/src/run_queue.rs
equation314: An init task is the task that is converted from the current execution flow, not switched from another task.

api/axfeat/Cargo.toml
hky1999: yep

modules/axtask/src/wait_queue.rs
equation314: I mean delete `TODO`, there is nothing to do.

modules/axtask/src/task.rs
hky1999: But you call `ret` in `context_switch`, after `ret`, the control flow jumps to `task_entry` directly (when it's the first time that the task runs)

hky1999: image
Single-step debugging

hky1999: image

hky1999: image

hky1999: image

The "new" task will start with `task_entry`, it will not reach here until it calls `yield_xxx`

equation314: Okay, you are right

modules/axtask/src/wait_queue.rs
hky1999: image
 Ok, how about `wait_time_until` ?

modules/axtask/src/wait_queue.rs
hky1999: Ok, the comments have been updated

modules/axtask/src/run_queue.rs
hky1999: Understand that

modules/axtask/src/task.rs
```diff
@@ -39,9 +50,26 @@ pub struct TaskInner {
...
     state: AtomicU8,
 
+    /// CPU affinity mask.
+    cpumask: SpinNoIrq<CpuMask>,
+
+    /// Mark whether the task is in the wait queue.
     in_wait_queue: AtomicBool,
+
+    /// Used to indicate whether the task is running on a CPU.
+    on_cpu: AtomicBool,
```
Azure-stars, line 58: Maybe it is unnecessary when `SMP` feature isn't enabled?

hky1999: I think it's better to add `#[cfg(feature = "smp")]` here, 
althrough polling on `on_cpu` will never happen in a singal core system

hky1999: @equation314

equation314: > I think it's better to add `#[cfg(feature = "smp")]` here, althrough polling on `on_cpu` will never happen in a singal core system

OK

modules/axtask/src/wait_queue.rs
equation314: Same

equation314: How about putting the guard into the loop?

hky1999: > How about putting the guard into the loop?

`set_alarm_wakeup` in `wait_time_until` use `TIMER_LIST.with_current` to manipulate timer_list, which is protected under `NoPreemptGuard`. 
But there will be no guard between `set_alarm_wakeup` and the loop bloc, is it ok to put the guard into the loop?

modules/axtask/src/wait_queue.rs
equation314: Too many `if`s, why not use the style before the modification?

modules/axtask/src/wait_queue.rs
```diff
@@ -27,49 +29,56 @@ use crate::{AxRunQueue, AxTaskRef, CurrentTask, RUN_QUEUE};
...
         }
     }
 
-    fn cancel_events(&self, curr: CurrentTask) {
+    /// Cancel events by removing the task from the wait queue.
+    /// If `from_timer_list` is true, the task should be removed from the timer list.
+    fn cancel_events(&self, curr: CurrentTask, from_timer_list: bool) {
         // A task can be wake up only one events (timer or `notify()`), remove
         // the event from another queue.
         if curr.in_wait_queue() {
```
guoweikang, line 57: 建议1：
之前我参考linux  在starry 那边 实现了一个watit_list   我映像arceos 应该也引入过类似的，有list其实可以解决 
-   task 里面  in_wait_list  状态的保存可以去掉了(我理解状态保存的原因是因为VecDeq 删除效率不高 )
-  删除效率有了保证

modules/axtask/src/run_queue.rs
```diff
@@ -75,54 +234,111 @@ impl AxRunQueue {
...
-        F: FnOnce(AxTaskRef),
-    {
+    /// Block the current task, put current task into the wait queue and reschedule.
+    /// Mark the state of current task as `Blocked`, set the `in_wait_queue` flag as true.
+    /// Note:
+    ///     1. The caller must hold the lock of the wait queue.
+    ///     2. The caller must ensure that the current task is in the running state.
+    ///     3. The caller must ensure that the current task is not the idle task.
+    ///     4. The lock of the wait queue will be released explicitly after current task is pushed into it.
+    pub fn blocked_resched(&mut self, mut wq_guard: WaitQueueGuard) {
```
guoweikang, line 289: 这个接口我认为不应该有 `wq_guard: WaitQueueGuard ` 这个入参，原因有几个
1、  `block` 应该是更宽泛的调度接口，而不仅仅是 `block on wait queue`
2、 `wait queue` 的相关动作(入队) 为什么不放在`waitqueue.rs` 里面实现呢？`lock guard` 跨文件和模块不是很好，而且这里确实可以放在它自己模块处理

modules/axtask/src/run_queue.rs
guoweikang, line 1: 建议2： blocked_resched这个接口我认为不应该有 wq_guard: WaitQueueGuard  这个入参，原因有几个
1、 block 应该是更宽泛的调度接口，考虑到以后更多可能的同步机制（其实timer 就是一种）， 而不仅仅是 block on wait queue
2、 wait queue 的相关动作(入队) 为什么不放在waitqueue.rs 里面实现呢？lock guard 跨文件和模块不是很好，而且这里确实可以放在它自己模块处理
其他接口也是类似,

guoweikang (commented):
1、on_cpu的设计是优雅的，我会尝试在starry借鉴一下
2、各个数据结构、模块之间的作用域可以更加界限分明一些，尽量避免跨域的`lock`
3、关于`list`使用我觉得可以直接复制`starry` 已有实现

modules/axtask/src/wait_queue.rs
hky1999: How about `wait` and `wait_until`, do they need `NoPreemptIrqSave` or just `NoPreempt` is fine

modules/axtask/src/run_queue.rs
hky1999: See this comment https://github.com/arceos-org/arceos/pull/176#discussion_r1788855228

 we have to make sure task's state is set as Blocked with holding the lock of wait queue.
 
 While @equation314 suggested that we should put all task state manipulation to `run_queue.rs`

Besides, instead of "block on wait queue`, where should a task be if it is not running or ready

modules/axtask/src/run_queue.rs
hky1999: see these two commits

[[refactor] pass SpinNoIrqGuard to blocked_resched](https://github.com/arceos-org/arceos/pull/176/commits/868ac3e3621e35c2b3821fa91dda076bff5a6712)
[868ac3e](https://github.com/arceos-org/arceos/pull/176/commits/868ac3e3621e35c2b3821fa91dda076bff5a6712)


[[refactor] merge unblock_task, add TaskUnblockGuard](https://github.com/arceos-org/arceos/pull/176/commits/8b1fe9b0f59e2671b14b6a34a32bb67b27ac2372)
[8b1fe9b](https://github.com/arceos-org/arceos/pull/176/commits/8b1fe9b0f59e2671b14b6a34a32bb67b27ac2372)

modules/axtask/src/run_queue.rs
guoweikang: **Regarding whether there will be other blocking situations**
I mean, are there other synchronization mechanisms (futex wait, completion wait,...) that are also blocking, not blocking on the waitqueue data structure, but blocking on some other queues
**Regarding the issue of whether the state needs to be locked**
I temporarily agree that the state machine settings need to be unified somewhere (whether it is really necessary, Linux seems to be more casual). The key purpose of doing this here is that the change and judgment of the state need to be protected by a lock at the same time. The logic of the wait queue should be reused here, so if the state change has its own lock, isn't it better?

modules/axtask/src/wait_queue.rs
hky1999: Ok, this will be improved in next PR.

In this PR we'll focus on the framework, make sure a bug-free per-cpu run queue

guoweikang: OK

modules/axtask/src/run_queue.rs
hky1999: 1. For arceos, futex wait, completion wait can all implemented based on wait queue (like Mutex in ArceOS's api).

2. modifications to the state may require locking at different granularities, such as from the scheduler or wait queue. Some state modifications can be lock-free, such as changes to the current task's state in `yield_current` or `sleep_until`.

guoweikang: Here is an implementation of a synchronization mechanism. I don’t quite remember why it was implemented this way, but what is certain is that waitque is definitely not a simple data structure with the smallest granularity.
https://github.com/Starry-OS/axsync.git ；
Of course, I won’t dwell on this too much, because it may not be wise to think too much about extensibility for features that don’t exist or haven’t been implemented yet; but I want everyone to know that this is possible.

hky1999: > but what is certain is that waitque is definitely not a simple data structure with the smallest granularity.
https://github.com/Starry-OS/axsync.git ；

So what is the difference between `WaitTaskList` and `WaitQueue` besides their data structure.

modules/axtask/src/run_queue.rs
```diff
@@ -75,54 +234,111 @@ impl AxRunQueue {
...
 
+    /// Unblock one task by inserting it into the run queue.
+    /// If task's `on_cpu` flag is true,
+    /// it will enter a loop until the task finishes its scheduling process,
+    /// see `unblock_locked` for details.
     pub fn unblock_task(&mut self, task: AxTaskRef, resched: bool) {
-        debug!("task unblock: {}", task.id_name());
+        // When irq is enabled, use `unblock_lock` to protect the task from being unblocked by timer and `notify()` at the same time.
+        #[cfg(feature = "irq")]
+        let _unblock_lock = task.get_unblock_lock();
```
guoweikang: 是不是可以不适用unblock lock，利用atomic_cmp?

guoweikang (approved):

modules/axtask/src/wait_queue.rs
```diff
@@ -148,6 +147,7 @@ impl WaitQueue {
 
         let mut timeout = true;
         while axhal::time::wall_time() < deadline {
+            let mut rq = current_run_queue::<NoPreemptIrqSave>();
```
equation314, line 150: What about the timer event triggering between `while ...` and `let mut rq = ...`? Then the task will block and the timeout will not take effect.

hky1999: you are right, it has been fixed.

modules/axtask/src/run_queue.rs
hky1999: updated in
[[refactor] delete unblock_locked, use atomic compare_exchange in unbl…](https://github.com/arceos-org/arceos/pull/176/commits/e64a60a16263728c0b53c7397532fc8a0756eaec) 
[e64a60a](https://github.com/arceos-org/arceos/pull/176/commits/e64a60a16263728c0b53c7397532fc8a0756eaec)

modules/axtask/src/wait_queue.rs
equation314: A better solution is to still keep the guard in the loop, but also move the deadline check in the guard: 

```rs
loop {
    let mut rq = current_run_queue::<NoPreemptIrqSave>();
    if axhal::time::wall_time() >= deadline {
        break;
    }
}
```

hky1999: Okay, it has been modified to 
```Rust
        loop {
            let mut rq = current_run_queue::<NoPreemptIrqSave>();
            if axhal::time::wall_time() >= deadline {
                break;
            }
            let wq = self.queue.lock();
            if condition() {
                timeout = false;
                break;
            }

            rq.blocked_resched(wq);
        }
```

hky1999:
Modifications to arceos-apps: https://github.com/arceos-org/arceos-apps/pull/1/files

modules/axtask/src/task.rs
```diff
@@ -199,6 +228,10 @@ impl TaskInner {
     pub(crate) fn new_init(name: String) -> Self {
         let mut t = Self::new_common(TaskId::new(), name);
         t.is_init = true;
+        #[cfg(feature = "smp")]
+        t.set_on_cpu(true);
+        // Init task can be scheduled on all CPUs.
+        t.set_cpumask(CpuMask::full());
```
equation314: We have already set the full cpumask in `new_common`.

modules/axtask/src/task.rs
```diff
@@ -320,6 +394,57 @@ impl TaskInner {
...
     }
 
+    /// Returns whether the task is running on a CPU.
+    ///
+    /// It is used to protect the task from being moved to a different run queue
+    /// while it has not finished its scheduling process.
+    /// The `on_cpu field is set to `true` when the task is preparing to run on a CPU,
+    /// and it is set to `false` when the task has finished its scheduling process in `clear_prev_task_on_cpu()`.
+    #[cfg(feature = "smp")]
+    pub fn on_cpu(&self) -> bool {
```
equation314: `pub(crate)`, we put all functions public to other crates in the first impl block.

Same as the following 3 functions.

modules/axtask/src/task.rs
```diff
@@ -320,6 +394,57 @@ impl TaskInner {
...
+    pub fn set_on_cpu(&self, on_cpu: bool) {
+        self.on_cpu.store(on_cpu, Ordering::Release)
+    }
+
+    /// Stores a weak reference to the previous task running on this CPU.
+    ///
+    /// ## Safety
+    /// This function is only called by current task in `switch_to`.
+    #[cfg(feature = "smp")]
+    pub unsafe fn set_prev_task(&self, prev_task: Arc<AxTask>) {
```
equation314: Change to `prev_task: &AxTaskRef` to avoid `Arc::clone`

modules/axtask/src/task.rs
```diff
@@ -320,6 +394,57 @@ impl TaskInner {
...
+    /// Panic if the pointer is invalid or the previous task is dropped.
+    ///
+    /// ## Note
+    /// This must be the very last reference to @_prev_task from this CPU.
+    /// After `on_cpu` is cleared, the task can be moved to a different CPU.
+    /// We must ensure this doesn't happen until the switch is completely finished.
+    ///
+    /// ## Safety
+    /// The caller must ensure that the weak reference to the prev task is valid, which is
+    /// done by the previous task running on this CPU through `set_prev_task_on_cpu_ptr()`.
```
equation314: Update comments

modules/axtask/src/task.rs
```diff
@@ -320,6 +394,57 @@ impl TaskInner {
...
+    #[cfg(feature = "smp")]
+    pub unsafe fn clear_prev_task_on_cpu(&self) {
+        self.prev_task
+            .get()
+            .as_ref()
+            .expect("Invalid prev_task pointer")
```
equation314: We can merge the two `expect`s:

```rs
self.prev_task
    .get()
    .as_ref()
    .and_then(Weak::upgrade)
    .expect("Invalid prev_task pointer")
    .set_on_cpu(false);
```

modules/axtask/src/wait_queue.rs
```diff
@@ -170,14 +183,13 @@ impl WaitQueue {
     /// preemption is enabled.
     pub fn notify_all(&self, resched: bool) {
         loop {
-            let mut rq = RUN_QUEUE.lock();
-            if let Some(task) = self.queue.lock().pop_front() {
-                task.set_in_wait_queue(false);
-                rq.unblock_task(task, resched);
+            let mut wq = self.queue.lock();
```
equation314: Reimplement it by `notify_one`?

modules/axtask/src/wait_queue.rs
```diff
@@ -27,49 +29,56 @@ use crate::{AxRunQueue, AxTaskRef, CurrentTask, RUN_QUEUE};
...
     /// Blocks the current task and put it into the wait queue, until other task
     /// notifies it.
     pub fn wait(&self) {
-        RUN_QUEUE.lock().block_current(|task| {
-            task.set_in_wait_queue(true);
-            self.queue.lock().push_back(task)
-        });
-        self.cancel_events(crate::current());
+        let mut rq = current_run_queue::<NoPreemptIrqSave>();
+        rq.blocked_resched(self.queue.lock());
```
equation314, line 81: Can we acquire `self.queue.lock` first, then call `current_run_queue` with `NoOp`, as `self.queue.lock` will disable IRQs?

modules/axtask/src/run_queue.rs
```diff
@@ -1,45 +1,202 @@
 use alloc::collections::VecDeque;
 use alloc::sync::Arc;
-use kspin::SpinNoIrq;
+use core::mem::MaybeUninit;
+#[cfg(feature = "smp")]
+use core::sync::atomic::{AtomicUsize, Ordering};
```
equation314: These types only used in `select_run_queue_index`, we can import them within the function.

modules/axtask/src/run_queue.rs
```diff
@@ -1,45 +1,202 @@
...
+    fn drop(&mut self) {
+        G::release(self.state);
     }
+}
 
+/// Core functions of run queue.
+impl<'a, G: BaseGuard> AxRunQueueRef<'a, G> {
     pub fn add_task(&mut self, task: AxTaskRef) {
-        debug!("task spawn: {}", task.id_name());
+        debug!("Add {} on run_queue {}", task.id_name(), self.inner.cpu_id);
```
equation314: The log style is inconsistent with other task ops, change to `task add: xxxxxx`

modules/axtask/src/run_queue.rs
```diff
@@ -75,54 +235,117 @@ impl AxRunQueue {
...
+        curr.set_in_wait_queue(true);
+
+        wq_guard.push_back(curr.clone());
+        // Drop the lock of wait queue explictly.
+        drop(wq_guard);
+
+        // Current task's state has been changed to `Blocked` and added to the wait queue.
+        // Note that the state may have been set as `Ready` in `unblock_task()`,
+        // see `unblock_task()` for details.
+        assert!(
```
equation314: This assertion seems obvious and can be deleted, will it be in the running or exited state?

modules/axtask/src/run_queue.rs
```diff
@@ -140,30 +363,62 @@ impl AxRunQueue {
...
-            self.resched(false);
+            self.inner.resched(false);
         }
     }
 }
 
 impl AxRunQueue {
+    /// Create a new run queue for the specified CPU.
+    /// The run queue is initialized with a per-CPU gc task in its scheduler.
+    pub fn new(cpu_id: usize) -> Self {
```
equation314: It can be a private fn

modules/axtask/src/run_queue.rs
```diff
@@ -75,54 +235,117 @@ impl AxRunQueue {
...
+    /// This function will never return.
     pub fn exit_current(&mut self, exit_code: i32) -> ! {
         let curr = crate::current();
         debug!("task exit: {}, exit_code={}", curr.id_name(), exit_code);
-        assert!(curr.is_running());
+        assert!(curr.is_running(), "task is not running: {:?}", curr.state());
         assert!(!curr.is_idle());
         if curr.is_init() {
-            EXITED_TASKS.lock().clear();
+            EXITED_TASKS.with_current(|exited_tasks| exited_tasks.clear());
```
equation314: It is called from `current_run_queue::<NoPreemptIrqSave>().exit_current(exit_code)`, which disabled preemption and IRQs, we can use `current_ref_raw`.

modules/axtask/src/run_queue.rs
```diff
@@ -176,65 +431,98 @@ impl AxRunQueue {
...
                 } else {
                     // Otherwise (e.g, `switch_to` is not compeleted, held by the
                     // joiner, etc), push it back and wait for them to drop first.
-                    EXITED_TASKS.lock().push_back(task);
+                    EXITED_TASKS.with_current(|exited_tasks| exited_tasks.push_back(task));
                 }
             }
         }
-        WAIT_FOR_EXIT.wait();
+        unsafe { WAIT_FOR_EXIT.current_ref_raw() }.wait();
```
equation314, line 492: Explain why use `current_ref_raw`

modules/axtask/src/run_queue.rs
```diff
@@ -176,65 +431,98 @@ impl AxRunQueue {
...
+    let cpu_id = this_cpu_id();
+
     // Put the subsequent execution into the `idle` task.
     let idle_task = TaskInner::new_init("idle".into()).into_arc();
     idle_task.set_state(TaskState::Running);
     IDLE_TASK.with_current(|i| {
         i.init_once(idle_task.clone());
     });
     unsafe { CurrentTask::init_current(idle_task) }
+    RUN_QUEUE.with_current(|rq| {
```
equation314, line 532: Add a blank line before it for better readability.

modules/axtask/src/task.rs
hky1999: modify to
```Rust
    #[cfg(feature = "smp")]
    pub(crate) unsafe fn clear_prev_task_on_cpu(&self) {
        self.prev_task
            .get()
            .as_ref()
            .and_then(|weak| weak.upgrade())
            .expect("Invalid prev_task pointer or prev_task has been dropped")
            .set_on_cpu(false);
    }
```

modules/axtask/src/wait_queue.rs
hky1999: I DO NOT think so.
the lock of wait queue has to be dropped before `resched`

while the guard of run queue will not be dropped until current task is back to running

modules/axtask/src/run_queue.rs
hky1999: Nope, since it is `blocked_resched` and it accepts a `WaitQueueGuard` as a parameter, it will set current task's state as `Blocked` with holding the `WaitQueueGuard`, insert current task into wq and then drop the `WaitQueueGuard`.

It's what happens after the dropping of `WaitQueueGuard` that matters, 

since then current task maybe unblocked by other cores and set its state as `Ready` due to `unblock_task()`'s `task.transition_state(TaskState::Blocked, TaskState::Ready)`

modules/axtask/src/run_queue.rs
hky1999: I think `with_current` with preemption disabled is safer here

modules/axtask/src/wait_queue.rs
hky1999: Reimplement as
```Rust
    /// Wakes all tasks in the wait queue.
    ///
    /// If `resched` is true, the current task will be preempted when the
    /// preemption is enabled.
    pub fn notify_all(&self, resched: bool) {
        loop {
            if self.queue.lock().is_empty() {
                break;
            }
            self.notify_one(resched);
        }
    }
```

modules/axtask/src/run_queue.rs
hky1999: Nononono, `current_ref_raw` is required here, we cannot block current task with preemption disabled, so we use `current_ref_raw` to get the `WAIT_FOR_EXIT`'s reference here to avoid the use of `NoPreemptGuard`.

hky1999: use `with_current` will lead to https://github.com/arceos-org/arceos/actions/runs/11287770863/job/31394266857?pr=176

hky1999: Do we need to add `kernel_guard::IrqSave` here?
@Azure-stars

hky1999: Since gc task is pinned to the current CPU, there is no affection if the gc task is preempted during the process.
So `kernel_guard::IrqSave` is not needed.

modules/axtask/src/wait_queue.rs
equation314: Can we use this:

```rs
while self.notify_one(resched) {}
```

hky1999: You are amazing
I want to marry you, bro

```Rust
    /// Wakes all tasks in the wait queue.
    ///
    /// If `resched` is true, the current task will be preempted when the
    /// preemption is enabled.
    pub fn notify_all(&self, resched: bool) {
        while self.notify_one(resched) {
            // loop until the wait queue is empty
        }
    }
```

modules/axtask/src/run_queue.rs
equation314: `assert(is_blocked || is_ready)` == `assert(is_not_running && is_not_exited)`，which is obvious.

BTW, `|` -> `||`

hky1999: I'll just delete the assertion but leave the comments

equation314 (approved):

PR #178, opened 2024-09-23
[mm] Add functions to read and write data in AddrSpace

Azure-stars:


modules/axmm/src/aspace.rs
```diff
@@ -109,6 +114,70 @@ impl AddrSpace {
...
+
+    /// To read data from the address space.
+    ///
+    /// # Arguments
+    ///
+    /// * `start` - The start virtual address to read.
+    /// * `buf` - The buffer to store the data. If the buffer is not large enough,
+    /// it will truncate the data to fit the buffer.
+    pub fn read(&self, start: VirtAddr, buf: &mut [u8]) -> AxResult {
+        if !self.contains_range(start, buf.len()) {
```
equation314: Move this line into `process_area_data`

equation314 (approved):

modules/axmm/src/aspace.rs
Azure-stars: fixed

PR #179, opened 2024-09-23
[hal] Support user space for riscv64 and aarch64

Azure-stars:


modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -15,6 +15,145 @@ pub struct TrapFrame {
     pub spsr: u64,
 }
 
+impl TrapFrame {
+    /// Gets the 0th syscall argument.
+    pub const fn arg0(&self) -> usize {
```
aarkegz, line 22: Can we just use `pub const fn arg(&self, index: usize) -> usize`?

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -15,6 +15,145 @@ pub struct TrapFrame {
...
+
+/// Context to enter user space.
+#[cfg(feature = "uspace")]
+pub struct UspaceContext(TrapFrame);
+
+#[cfg(feature = "uspace")]
+impl UspaceContext {
+    /// Creates an empty context with all registers set to zero.
+    pub const fn empty() -> Self {
+        unsafe { core::mem::MaybeUninit::zeroed().assume_init() }
```
aarkegz, line 60: Maybe `Default::default()`?

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -15,6 +15,145 @@ pub struct TrapFrame {
...
+
+    /// Gets the 5th syscall argument.
+    pub const fn arg5(&self) -> usize {
+        self.r[5] as _
+    }
+}
+
+/// Context to enter user space.
+#[cfg(feature = "uspace")]
+pub struct UspaceContext(TrapFrame);
```
aarkegz, line 54: I think `UserSpaceContext` is a better name.

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -15,6 +15,145 @@ pub struct TrapFrame {
...
+        self.0.elr as _
+    }
+
+    /// Gets the stack pointer.
+    pub const fn get_sp(&self) -> usize {
+        self.0.usp as _
+    }
+
+    /// Sets the instruction pointer.
+    pub const fn set_ip(&mut self, pc: usize) {
```
aarkegz, line 98: `ip` or `pc`?

modules/axhal/src/arch/riscv/context.rs
```diff
@@ -53,6 +53,136 @@ pub struct TrapFrame {
     pub sstatus: usize,
 }
 
+impl TrapFrame {
+    /// Gets the 0th syscall argument.
+    pub const fn arg0(&self) -> usize {
```
aarkegz, line 60: Same as aarch64.

modules/axhal/src/arch/aarch64/context.rs
equation314: It is not a const fn

modules/axhal/src/arch/aarch64/context.rs
equation314: It may introduce overhead of index matching.

modules/axhal/src/arch/aarch64/context.rs
aarkegz: That makes sense

modules/axhal/src/arch/aarch64/context.rs
equation314: `set_ip` is a generic interface for multiple architectures, and should be given a uniform name to have consistent calling code。 As for why we don't use `set_pc`, because `ip` corresponds to `sp`.

`pc` is a parameter whose naming does not affect the call and can have an architecture specific name (`rip` for x86_64, `pc` for aarch64 and riscv64).

modules/axhal/src/arch/aarch64/context.rs
equation314: Maybe, but my reason is that `UspaceContext` corresponds to the `uspace` feature.

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -15,6 +15,145 @@ pub struct TrapFrame {
...
+    }
+
+    /// Sets the stack pointer.
+    pub const fn set_sp(&mut self, sp: usize) {
+        self.0.usp = sp as _;
+    }
+
+    /// Sets the return value register.
+    pub const fn set_retval(&mut self, retval: usize) {
+        self.0.r[8] = retval as _;
```
equation314: `r[0]`

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -15,6 +15,145 @@ pub struct TrapFrame {
...
+    }
+
+    /// Enters user space.
+    ///
+    /// It restores the user registers and jumps to the user entry point
+    /// (saved in `sepc`).
```
equation314:
```
(saved in `elr`)
```

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -15,6 +15,145 @@ pub struct TrapFrame {
...
+        self.0.elr = pc as _;
+    }
+
+    /// Sets the stack pointer.
+    pub const fn set_sp(&mut self, sp: usize) {
+        self.0.usp = sp as _;
+    }
+
+    /// Sets the return value register.
+    pub const fn set_retval(&mut self, retval: usize) {
```
equation314: `retval` -> `r0`

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -15,6 +15,145 @@ pub struct TrapFrame {
...
+    /// switched to `kstack_top`.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it changes processor mode and the stack.
+    #[inline(never)]
+    #[no_mangle]
+    pub unsafe fn enter_uspace(&self, kstack_top: VirtAddr) -> ! {
+        super::disable_irqs();
+        // We don't process elxt trap, so we don't need to save the current kstack ptr to the taskctx.
```
equation314: What is `elxt`?

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -75,25 +217,48 @@ impl TaskContext {
     ///
     /// [`init`]: TaskContext::init
     /// [`switch_to`]: TaskContext::switch_to
-    pub const fn new() -> Self {
-        unsafe { core::mem::MaybeUninit::zeroed().assume_init() }
+    pub fn new() -> Self {
+        Self {
+            #[cfg(feature = "uspace")]
+            ttbr0_el1: crate::paging::kernel_page_table_root(),
```
equation314: The page table setting is a bit different between AArch64 and x86_64.

The default task context is used for a kernel thread, it use `TTBR1` rather than `TTBR0` for page table root, so we need to keep `ttbr0_el1` zero.

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -75,25 +217,48 @@ impl TaskContext {
...
     pub fn init(&mut self, entry: usize, kstack_top: VirtAddr, tls_area: VirtAddr) {
         self.sp = kstack_top.as_usize() as u64;
         self.lr = entry as u64;
+        // When under `uspace` feature, kernel will not use this register.
         self.tpidr_el0 = tls_area.as_usize() as u64;
     }
 
+    /// Changes the page table root for user space (`ttbr0_el1` register for aarch64 in el1 level).
+    ///
+    /// If not set, the kernel page table root is used (obtained by
```
equation314: Change comments to describe the differences with x86.

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -75,25 +217,48 @@ impl TaskContext {
...
+    }
+
     /// Switches to another task.
     ///
     /// It first saves the current task's context from CPU to this place, and then
     /// restores the next task's context from `next_ctx` to CPU.
     pub fn switch_to(&mut self, next_ctx: &Self) {
         #[cfg(feature = "fp_simd")]
         self.fp_state.switch_to(&next_ctx.fp_state);
+
```
equation314: Remove the empty line.

modules/axhal/src/arch/aarch64/mod.rs
```diff
@@ -4,6 +4,8 @@ pub(crate) mod trap;
 use core::arch::asm;
 
 use aarch64_cpu::registers::{DAIF, TPIDR_EL0, TTBR0_EL1, TTBR1_EL1, VBAR_EL1};
+#[cfg(feature = "uspace")]
+pub use context::UspaceContext;
```
equation314: Put this line before line 12 and change to `pub use self::context::UspaceContext`.

modules/axhal/src/arch/riscv/trap.rs
```diff
@@ -37,6 +37,11 @@ fn handle_page_fault(tf: &TrapFrame, mut access_flags: MappingFlags, is_user: bo
 fn riscv_trap_handler(tf: &mut TrapFrame, from_user: bool) {
     let scause = scause::read();
     match scause.cause() {
+        #[cfg(feature = "uspace")]
+        Trap::Exception(E::UserEnvCall) => {
+            tf.regs.a0 = crate::trap::handle_syscall(tf, tf.regs.a7 as usize) as usize;
```
equation314: Remove the first `as usize`

modules/axhal/src/arch/riscv/context.rs
```diff
@@ -53,6 +53,136 @@ pub struct TrapFrame {
...
+        self.regs.a0 as _
```
equation314: Remove `as _`

modules/axhal/src/arch/riscv/context.rs
```diff
@@ -53,6 +53,136 @@ pub struct TrapFrame {
...
+    }
+
+    /// Creates a new context from the given [`TrapFrame`].
+    pub const fn from(trap_frame: &TrapFrame) -> Self {
+        Self(*trap_frame)
+    }
+
+    /// Gets the instruction pointer.
+    pub const fn get_ip(&self) -> usize {
+        self.0.sepc as _
```
equation314: Remove `as _`

modules/axhal/src/arch/riscv/context.rs
```diff
@@ -53,6 +53,136 @@ pub struct TrapFrame {
...
+        self.0.sepc = pc as _;
+    }
+
+    /// Sets the stack pointer.
+    pub const fn set_sp(&mut self, sp: usize) {
+        self.0.regs.sp = sp as _;
+    }
+
+    /// Sets the return value register.
+    pub const fn set_retval(&mut self, retval: usize) {
```
equation314: `retval` -> `a0`

modules/axhal/src/arch/riscv/context.rs
```diff
@@ -53,6 +53,136 @@ pub struct TrapFrame {
...
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it changes processor mode and the stack.
+    #[inline(never)]
+    #[no_mangle]
+    pub unsafe fn enter_uspace(&self, kstack_top: VirtAddr) -> ! {
+        super::disable_irqs();
+
+        use riscv::register::{sepc, sscratch};
```
equation314, line 160: Put `use` at the beginning of the function.

modules/axhal/src/arch/riscv/context.rs
```diff
@@ -109,16 +246,33 @@ impl TaskContext {
...
+        self.satp = satp;
+    }
+
     /// Switches to another task.
     ///
     /// It first saves the current task's context from CPU to this place, and then
     /// restores the next task's context from `next_ctx` to CPU.
     pub fn switch_to(&mut self, next_ctx: &Self) {
-        #[cfg(feature = "tls")]
+        #[cfg(any(feature = "tls", feature = "uspace"))]
```
equation314: Remove this change

modules/axhal/src/arch/aarch64/context.rs
Azure-stars: Here x is the currently running exception level, which means the trap that occurs at the current privilege level.

hky1999: You can use ELx to get rid of unnecessary confusion

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -1,3 +1,4 @@
+use aarch64_cpu::registers::SPSR_EL1;
```
equation314: Move it into `UspaceContext::new`, otherwise it will cause a unused import warning.

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -15,6 +16,146 @@ pub struct TrapFrame {
...
+    /// switched to `kstack_top`.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it changes processor mode and the stack.
+    #[inline(never)]
+    #[no_mangle]
+    pub unsafe fn enter_uspace(&self, kstack_top: VirtAddr) -> ! {
+        super::disable_irqs();
+        // We do not handle traps that occur at the current exception level,
```
equation314, line 126: The exact reason is that we set `SPSel` = `ELx`, so that different `SP` is used at different exception levels.

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -65,6 +206,9 @@ pub struct TaskContext {
     pub lr: u64, // r30
     #[cfg(feature = "fp_simd")]
     pub fp_state: FpState,
+    /// The `ttbr0_el1` register value, i.e., the page table root.
+    #[cfg(feature = "uspace")]
+    pub ttbr0_el1: memory_addr::PhysAddr,
```
equation314: Put it before `fp_state` to make the structure layout compact.

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -75,25 +219,45 @@ impl TaskContext {
     ///
     /// [`init`]: TaskContext::init
     /// [`switch_to`]: TaskContext::switch_to
-    pub const fn new() -> Self {
-        unsafe { core::mem::MaybeUninit::zeroed().assume_init() }
+    pub fn new() -> Self {
```
equation314, line 224: It can be written as `Self::default()`

modules/axhal/src/arch/riscv/trap.S
```diff
@@ -10,18 +10,22 @@
     STR     t2, sp, 1                   // tf.regs.sp
 
 .if \from_user == 1
+    LDR     t1, sp, 2                   // load supervisor gp
     LDR     t0, sp, 3                   // load supervisor tp
     STR     gp, sp, 2                   // save user gp and tp
     STR     tp, sp, 3
     mv      tp, t0
+    mv      gp, t1
```
equation314: Keep the order of `gp` and `tp` consistent.

modules/axhal/src/arch/x86_64/context.rs
```diff
@@ -280,6 +280,9 @@ pub struct TaskContext {
     /// The `CR3` register value, i.e., the page table root.
     #[cfg(feature = "uspace")]
     pub cr3: PhysAddr,
+    /// The `gs_base` register value.
+    #[cfg(feature = "uspace")]
+    pub gs_base: usize,
```
equation314: Put it after `fs_base`

modules/axhal/src/arch/aarch64/trap.S
```diff
@@ -21,6 +21,13 @@
     mrs     x11, spsr_el1
     stp     x30, x9, [sp, 30 * 8]
     stp     x10, x11, [sp, 32 * 8]
+    /*
+    * We may have interrupted userspace, or a guest, or exit-from or 
+    * return-to either of those. So we can't trust sp_el0, and need to
+    * restore it.
+     */
```
equation314: Format comments

modules/axhal/src/cpu.rs
```diff
@@ -22,6 +22,14 @@ pub fn this_cpu_is_bsp() -> bool {
     IS_BSP.read_current()
 }
 
+/// Gets the pointer to the current task without preemption-safety.
+///
+/// Preemption and irq must be disabled when calling this function.
+#[no_mangle]
+unsafe fn current_task_ptr_raw() -> usize {
```
equation314: Rename to `cache_current_task_ptr` and update `SP_EL0` here.

modules/axhal/src/cpu.rs
```diff
@@ -68,8 +77,10 @@ pub unsafe fn set_current_task_ptr<T>(ptr: *const T) {
     }
     #[cfg(target_arch = "aarch64")]
     {
+        let _guard = kernel_guard::IrqSave::new();
+        CURRENT_TASK_PTR.write_current_raw(ptr as usize);
         use tock_registers::interfaces::Writeable;
-        aarch64_cpu::registers::SP_EL0.set(ptr as u64)
+        aarch64_cpu::registers::SP_EL0.set(ptr as u64);
```
equation314:
```rs
let _guard = kernel_guard::IrqSave::new();
CURRENT_TASK_PTR.write_current_raw(ptr as usize);
cache_current_task_ptr();
```

modules/axhal/src/cpu.rs
```diff
@@ -22,6 +22,14 @@ pub fn this_cpu_is_bsp() -> bool {
     IS_BSP.read_current()
 }
 
+/// Gets the pointer to the current task without preemption-safety.
+///
+/// Preemption and irq must be disabled when calling this function.
+#[no_mangle]
```
equation314: Remove `#[no_mangle]` and pass `xxx = sym xxx` in `global_asm!(include_str!(“trap.S”))`.

modules/axhal/src/cpu.rs
```diff
@@ -22,6 +22,15 @@ pub fn this_cpu_is_bsp() -> bool {
     IS_BSP.read_current()
 }
 
+/// Gets the pointer to the current task without preemption-safety.
```
equation314: Fix comments

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -15,6 +15,147 @@ pub struct TrapFrame {
     pub spsr: u64,
 }
 
+impl TrapFrame {
```
hky1999, line 20: can this be generate through macro like

`impl_argx!{}`

modules/axhal/src/arch/aarch64/context.rs
```diff
@@ -75,25 +219,40 @@ impl TaskContext {
...
         self.lr = entry as u64;
+        // When under `uspace` feature, kernel will not use this register.
         self.tpidr_el0 = tls_area.as_usize() as u64;
     }
 
+    /// Changes the page table root for user space (`ttbr0_el1` register for aarch64 in el1 level).
+    ///
+    /// If not set, it means that this task is a kernel task and only `ttbr1_el1` register will be used.
+    #[cfg(feature = "uspace")]
+    pub fn set_page_table_root(&mut self, ttbr0_el1: memory_addr::PhysAddr) {
```
hky1999: add `import memory_addr::PhysAddr` in the beginning of this file.
avoid to use it as the type definition in function arguments

modules/axhal/src/arch/aarch64/trap.S
```diff
@@ -21,6 +21,11 @@
     mrs     x11, spsr_el1
     stp     x30, x9, [sp, 30 * 8]
     stp     x10, x11, [sp, 32 * 8]
+    
+    # We may have interrupted userspace, or a guest, or exit-from or 
+    # return-to either of those. So we can't trust sp_el0, and need to
+    # restore it.
+    bl      {cache_current_task_ptr}
```
hky1999, line 28: you can add `sp_el0` explicitl in thic function name, to claim that we use `sp_el0` to store current task ptr
BTW, is it a caching operation??

modules/axhal/src/arch/aarch64/trap.rs
```diff
@@ -1,12 +1,11 @@
 use core::arch::global_asm;
 
+use super::TrapFrame;
```
hky1999: move `super::TrapFrame` to line 7.

Typically, references from this crate should be placed at the end.

modules/axhal/src/arch/riscv/context.rs
```diff
@@ -53,6 +53,136 @@ pub struct TrapFrame {
     pub sstatus: usize,
 }
 
+impl TrapFrame {
```
hky1999, line 58: I really think this should be written as macros
some thing like 
```Rust
macro_rules! define_syscall_args {
    ($struct_name:ident, $( $index:expr ),*) => {
        impl $struct_name {
            $( 
                /// Gets the syscall argument.
                pub const fn arg$index(&self) -> usize {
                    self.regs.a$index
                }
            )*
        }
    };
}

// Usage
define_syscall_args!(TrapFrame, 0, 1, 2, 3, 4, 5);
```

modules/axhal/src/arch/riscv/context.rs
hky1999: There are similar issues in aarch64 related files ; it's best to place these references at the start of the file.

modules/axhal/src/arch/riscv/context.rs
aarkegz: This won't work due to rust's macro rules. To create a identifier by concatenation, we must use a proc_macro like `paste::paste!`, `arg$index` just won't work. But I do agree that we need a better way to declare syscall arg accessors.

equation314: We can do this without separating the register names in the middle：

```rs
define_syscall_args!(TrapFrame, rdi, rsi, rdx, r10, r8, r9); // x86_64
define_syscall_args!(TrapFrame, regs.a0, regs.a1, regs.a2, regs.a3, regs.a4, regs.a5); // riscv64
define_syscall_args!(TrapFrame, r[0], r[1], r[2], r[3], r[4], r[5]); // aarch64
```

modules/axhal/src/arch/aarch64/trap.S
Azure-stars: In ArceOS's aarch64 architecture, we use `sp_el0` as the read cache for the current task pointer. And this function `cache_current_task_ptr` will update this cache.

modules/axhal/src/arch/aarch64/trap.rs
Azure-stars: Thanks.I learn a lot.

modules/axhal/src/arch/aarch64/context.rs
Azure-stars: Done

modules/axhal/src/arch/riscv/context.rs
```diff
@@ -1,12 +1,13 @@
+#[allow(unused_imports)]
```
equation314: Add `!` after `#` and a newline after it.

modules/axhal/src/arch/riscv/context.rs
equation314: It may be in the next PR, let's merge this first.

equation314 (approved):

PR #184, opened 2024-10-16
Replace the wait_queue data structure VecQueue with a linked list

guoweikang:
--------
TODO: Replace branch with the official version number

### DIFF
--- a/api/arceos_posix_api/build.rs
+++ b/api/arceos_posix_api/build.rs
@@ -5,9 +5,9 @@ fn main() {
         // TODO: generate size and initial content automatically.
         let (mutex_size, mutex_init) = if cfg!(feature = "multitask") {
             if cfg!(feature = "smp") {
-                (6, "{0, 0, 8, 0, 0, 0}") // core::mem::transmute::<_, [usize; 6]>(axsync::Mutex::new(()))
+                (3, "{0, 0, 0}") // core::mem::transmute::<_, [usize; 3]>(axsync::Mutex::new(()))
             } else {
-                (5, "{0, 8, 0, 0, 0}") // core::mem::transmute::<_, [usize; 5]>(axsync::Mutex::new(()))
+                (2, "{0, 0}") // core::mem::transmute::<_, [usize; 2]>(axsync::Mutex::new(()))
             }
         } else {
             (1, "{0}")
--- a/modules/axtask/src/run_queue.rs
+++ b/modules/axtask/src/run_queue.rs
@@ -10,7 +10,7 @@ use scheduler::BaseScheduler;
 use axhal::cpu::this_cpu_id;
 
 use crate::task::{CurrentTask, TaskState};
-use crate::wait_queue::WaitQueueGuard;
+use crate::wait_queue::{WaitQueueGuard, WaitTaskNode};
 use crate::{AxCpuMask, AxTaskRef, Scheduler, TaskInner, WaitQueue};
 
 macro_rules! percpu_static {
@@ -367,7 +367,7 @@ impl<'a, G: BaseGuard> CurrentRunQueueRef<'a, G> {
     ///     2. The caller must ensure that the current task is in the running state.
     ///     3. The caller must ensure that the current task is not the idle task.
     ///     4. The lock of the wait queue will be released explicitly after current task is pushed into it.
-    pub fn blocked_resched(&mut self, mut wq_guard: WaitQueueGuard) {
+    pub fn blocked_resched(&mut self, mut wq_guard: WaitQueueGuard, curr_waiter: &WaitTaskNode) {
         let curr = &self.current_task;
         assert!(curr.is_running());
         assert!(!curr.is_idle());
@@ -380,9 +380,11 @@ impl<'a, G: BaseGuard> CurrentRunQueueRef<'a, G> {
         // Mark the task as blocked, this has to be done before adding it to the wait queue
         // while holding the lock of the wait queue.
         curr.set_state(TaskState::Blocked);
-        curr.set_in_wait_queue(true);
-
-        wq_guard.push_back(curr.clone());
+        // SAFETY: The waiter was on caller stack, the lifetime ends
+        // only when the task resumes running.
+        unsafe {
+            assert!(wq_guard.push_back(curr_waiter));
+        }
         // Drop the lock of wait queue explictly.
         drop(wq_guard);
 
--- a/modules/axtask/src/wait_queue.rs
+++ b/modules/axtask/src/wait_queue.rs
@@ -1,10 +1,20 @@
-use alloc::collections::VecDeque;
 use alloc::sync::Arc;
 
 use kernel_guard::{NoOp, NoPreemptIrqSave};
 use kspin::{SpinNoIrq, SpinNoIrqGuard};
+use linked_list_r4l::{def_node, RawList};
 
-use crate::{current_run_queue, select_run_queue, AxTaskRef, CurrentTask};
+use crate::{current_run_queue, select_run_queue, AxTaskRef};
+
+def_node! {
+    pub(crate) struct WaitTaskNode(AxTaskRef);
+}
+
+impl WaitTaskNode {
+    fn from_current() -> Self {
+        WaitTaskNode::new(crate::current().clone())
+    }
+}
 
 /// A queue to store sleeping tasks.
 ///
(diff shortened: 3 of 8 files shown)


modules/axtask/src/wait_queue.rs
```diff
@@ -1,10 +1,19 @@
...
-use crate::{current_run_queue, select_run_queue, AxTaskRef, CurrentTask};
+use crate::wait_list::{WaitTaskList, WaitTaskNode};
+use crate::{current_run_queue, select_run_queue, AxTaskRef};
+
+#[cfg(feature = "irq")]
+use crate::CurrentTask;
+
+macro_rules! declare_current_waiter {
+    ($name: ident) => {
+        let $name = Arc::new(WaitTaskNode::new($crate::current().as_task_ref().clone()));
```
hky1999: crate::current().clone()

modules/axtask/src/wait_queue.rs
```diff
@@ -29,57 +38,41 @@ use crate::{current_run_queue, select_run_queue, AxTaskRef, CurrentTask};
 /// assert_eq!(VALUE.load(Ordering::Relaxed), 1);
 /// ```
 pub struct WaitQueue {
-    queue: SpinNoIrq<VecDeque<AxTaskRef>>,
+    queue: SpinNoIrq<WaitTaskList>,
```
hky1999: Why do we wrap `WaitTaskList` with `WaitQueue`,
instead of use `list: SpinNoIrq<List<Node>>` here directly

modules/axtask/src/wait_queue.rs
```diff
@@ -117,12 +116,14 @@ impl WaitQueue {
...
+        rq.blocked_resched(self.queue.lock(), waiter.clone());
 
-        let timeout = curr.in_wait_queue(); // still in the wait queue, must have timed out
+        let timeout = axhal::time::wall_time() >= deadline;
 
+        // Always try to remove the task from wait list.
+        self.queue.lock().remove(&waiter);
         // Always try to remove the task from the timer list.
-        self.cancel_events(curr, true);
+        self.cancel_timer(curr);
```
hky1999: can we keep the `cancel_events` and move `self.queue.lock().remove(&waiter);` into it

modules/axtask/src/wait_queue.rs
```diff
@@ -117,12 +116,14 @@ impl WaitQueue {
         );
         crate::timers::set_alarm_wakeup(deadline, curr.clone());
 
-        rq.blocked_resched(self.queue.lock());
+        rq.blocked_resched(self.queue.lock(), waiter.clone());
 
-        let timeout = curr.in_wait_queue(); // still in the wait queue, must have timed out
+        let timeout = axhal::time::wall_time() >= deadline;
```
hky1999, line 127: we also need to delete `in_wait_queue` field and related functions from axtask

modules/axtask/src/wait_queue.rs
guoweikang: OK goog idea , it have done

modules/axtask/src/wait_queue.rs
guoweikang: It already deleted ,may be you didn't notice

modules/axtask/src/wait_queue.rs
guoweikang: good idea,  to previous version.

modules/axtask/src/wait_queue.rs
guoweikang: done

modules/axtask/src/run_queue.rs
```diff
@@ -286,7 +286,11 @@ impl<'a, G: BaseGuard> AxRunQueueRef<'a, G> {
     ///     2. The caller must ensure that the current task is in the running state.
     ///     3. The caller must ensure that the current task is not the idle task.
     ///     4. The lock of the wait queue will be released explicitly after current task is pushed into it.
-    pub fn blocked_resched(&mut self, mut wq_guard: WaitQueueGuard) {
+    pub fn blocked_resched(
+        &mut self,
+        mut wq_guard: WaitQueueGuard,
+        curr_waiter: Arc<WaitTaskNode>,
```
hky1999: It still seems awkward, `curr_waiter` is just a wrapper for `crate::current()` in this code block.

guoweikang: Here need param `waiter`, because wq_guard  have to  push back the `curr wrapper` , we can't crteate the wrapper in `blocked_resched`, because `waiter` fast delete in caller,  need use it; we can discuss  whether there is a better way

guoweikang: https://github.com/arceos-org/arceos/pull/184/commits/6219d10f9339aad5770d7a371294b72155d3a560   in this commit, give a plan  remove wq_guard and waiter

api/arceos_posix_api/build.rs
```diff
@@ -5,9 +5,9 @@ fn main() {
         // TODO: generate size and initial content automatically.
         let (mutex_size, mutex_init) = if cfg!(feature = "multitask") {
             if cfg!(feature = "smp") {
-                (6, "{0, 0, 8, 0, 0, 0}") // core::mem::transmute::<_, [usize; 6]>(axsync::Mutex::new(()))
+                (3, "{0, 0, 0}") // core::mem::transmute::<_, [usize; 6]>(axsync::Mutex::new(()))
```
equation314: Fix comments

modules/axtask/src/wait_queue.rs
```diff
@@ -188,24 +202,30 @@ impl WaitQueue {
...
+                        unblock_one_task(task.clone(), resched);
+                        break true;
+                    }
+                }
+                None => break false,
+            }
+            cursor.move_next();
         }
-    }
+    }
```
equation314: Remove space

modules/axtask/src/wait_queue.rs
```diff
@@ -135,50 +149,50 @@ impl WaitQueue {
...
     ///
     /// If `resched` is true, the current task will be preempted when the
     /// preemption is enabled.
     pub fn notify_one(&self, resched: bool) -> bool {
-        let mut wq = self.queue.lock();
-        if let Some(task) = wq.pop_front() {
-            unblock_one_task(task, resched);
+        let mut wq = self.list.lock();
+        if let Some(waiter) = wq.pop_front() {
+            unblock_one_task(waiter.inner().clone(), resched);
```
equation314: Add an `into_inner` method to the node type to avoid `Arc::clone`.

modules/axtask/src/wait_queue.rs
```diff
@@ -29,42 +40,46 @@ use crate::{current_run_queue, select_run_queue, AxTaskRef, CurrentTask};
...
+        let mut wq_locked = self.list.lock();
+        debug!("task block: {}", curr.id_name());
+        // Mark the task as blocked, this has to be done before adding it to the wait queue
+        // while holding the lock of the wait queue.
+        curr.set_state(TaskState::Blocked);
+        wq_locked.push_back(waiter.clone());
     }
 
-    /// Cancel events by removing the task from the wait queue.
+    /// Cancel events by removing the task from the wait list.
```
equation314: It's fine to call it a **queue** as the term “wait queue” is a proper noun. And the task is also FIFO when call wait and notify_one.

modules/axtask/src/wait_queue.rs
```diff
@@ -73,14 +88,16 @@ impl WaitQueue {
         }
     }
 
-    /// Blocks the current task and put it into the wait queue, until other task
+    /// Blocks the current task and put it into the wait list, until other task
     /// notifies it.
     pub fn wait(&self) {
-        current_run_queue::<NoPreemptIrqSave>().blocked_resched(self.queue.lock());
-        self.cancel_events(crate::current(), false);
+        declare_current_waiter!(waiter);
```
equation314: Why we need a macro? Is the following code OK?
```rs
let waiter = WaitTaskNode::from_current();
```

P.S. we can add methods in `WaitTaskNode` as we define it in this file.

modules/axtask/src/wait_queue.rs
guoweikang: In fact, we cannot use into_inner directly here, because there is a waiter reference in the blocked task context, and the waiter cannot be dropped directly.

modules/axtask/src/wait_queue.rs
guoweikang: Done

modules/axtask/src/wait_queue.rs
guoweikang: Recover

modules/axtask/src/wait_queue.rs
guoweikang: I used `into_inner` in `cancel_events` to ensure that the reference count is equal to 1 when the waiter is destroyed.

modules/axtask/src/wait_queue.rs
guoweikang: fixed

api/arceos_posix_api/build.rs
guoweikang: fixed

Issue #169, opened 2024-08-27
Bug: cannot run `cargo any-command --all-features` with `log` version `0.4.22`


aarkegz:
`log` version `0.4.22` will produce a compiler error if multiple `max_level_*` features are specified. As corresponding features exists in `axfeat`, `axstd` and `axlog`, `cargo any-command --all-features` will result in unavoidable compiler error.

Workarounds:
- Lock `log` to `0.4.21`, or
- `--exclude axlog --exclude axstd --exclude axfeat`.

aarkegz:
Locked in #189.

zjp-CN:
The breakage of multiple features on log max level was introduced on purpose since https://github.com/rust-lang/log/pull/627 . 
And some backgrounds on the behavior change: https://github.com/rust-lang/log/issues/481 , https://github.com/rust-lang/log/issues/635 .

Solutions I see for the upstream log crate
* Log crate maintainers wish to replace cargo features by environment variables + build.rs + cfg flags, but no plan in near future.
* [Mutually-excusive, global features](https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618) should solve the problem, but it's in the early pre-rfc stage.

Anyways, this issue has been addressed in https://github.com/arceos-org/arceos/pull/189 by locking the version of log. The issue should be closed.

equation314:
#189 is a temporary fix, this issue is not perfectly resolved yet. I think we should keep it opened.

PR #189, opened 2024-10-19
Use `=` syntax in Cargo.toml to lock `log` to `0.4.21`, see #169

aarkegz:


aarkegz:
Cargo.lock can protect us for now, however, if we want to run `cargo update` in the future, it could be a problem. So I think we can use `log = "=0.4.21"` to express this limitation explicitly.

equation314 (approved):

PR #191, opened 2024-10-22
Aarch64 Add Linux Image Header 

ZR233:
So that can be load by uboot easily.


equation314:
Could you give the reference to the documentation on the header format?

And why not use the `mkimage` tool to convert the binary to U-Boot image?

ZR233:
https://docs.kernel.org/arch/arm64/booting.html
`mkimage` uses info in this header to gen uimage.

ZR233:
change to  #221

PR #195, opened 2024-10-24
[feat] replace deprecated sbi_rt legacy console operations in riscv

hky1999:
found by https://github.com/arceos-hypervisor/riscv_vcpu/pull/8

which caused https://github.com/arceos-hypervisor/arceos-umhv/actions/runs/11483702522/job/31959762361 to fail

[console_putchar](https://github.com/rustsbi/rustsbi/blob/4036e1d5abb5f8ceb9fde28658205450f685b4f8/sbi-rt/src/legacy.rs#L31) and [console_getchar](https://github.com/rustsbi/rustsbi/blob/4036e1d5abb5f8ceb9fde28658205450f685b4f8/sbi-rt/src/legacy.rs#L24)
are deprecated

### DIFF
--- a/modules/axhal/src/platform/aarch64_bsta1000b/dw_apb_uart.rs
+++ b/modules/axhal/src/platform/aarch64_bsta1000b/dw_apb_uart.rs
@@ -21,6 +21,28 @@ pub fn putchar(c: u8) {
     }
 }
 
+/// Write a slice of bytes to the console.
+pub fn write_bytes(bytes: &[u8]) {
+    for c in bytes {
+        putchar(*c);
+    }
+}
+
+/// Reads bytes from the console into the given mutable slice.
+/// Returns the number of bytes read.
+pub fn read_bytes(bytes: &mut [u8]) -> usize {
+    let mut read_len = 0;
+    while read_len < bytes.len() {
+        if let Some(c) = getchar() {
+            bytes[read_len] = c;
+        } else {
+            break;
+        }
+        read_len += 1;
+    }
+    read_len
+}
+
 /// Reads a byte from the console, or returns [`None`] if no input is available.
 pub fn getchar() -> Option<u8> {
     UART.lock().getchar()
--- a/modules/axhal/src/platform/aarch64_bsta1000b/dw_apb_uart.rs
+++ b/modules/axhal/src/platform/aarch64_bsta1000b/dw_apb_uart.rs
@@ -21,6 +21,11 @@ pub fn putchar(c: u8) {
     }
 }
 
+/// Reads a byte from the console, or returns [`None`] if no input is available.
+fn getchar() -> Option<u8> {
+    UART.lock().getchar()
+}
+
 /// Write a slice of bytes to the console.
 pub fn write_bytes(bytes: &[u8]) {
     for c in bytes {
@@ -41,11 +46,6 @@ pub fn read_bytes(bytes: &mut [u8]) -> usize {
         read_len += 1;
     }
     read_len
-}
-
-/// Reads a byte from the console, or returns [`None`] if no input is available.
-pub fn getchar() -> Option<u8> {
-    UART.lock().getchar()
 }
 
 /// UART simply initialize
--- a/modules/axhal/src/platform/dummy/mod.rs
+++ b/modules/axhal/src/platform/dummy/mod.rs
@@ -11,6 +11,17 @@ pub mod console {
     pub fn getchar() -> Option<u8> {
         unimplemented!()
     }
+
+    /// Writes bytes to the console from input u8 slice.
+    pub fn write_bytes(_bytes: &[u8]) {
+        unimplemented!()
+    }
+
+    /// Reads bytes from the console into the given mutable slice.
+    /// Returns the number of bytes read.
+    pub fn read_bytes(_bytes: &mut [u8]) -> usize {
+        unimplemented!()
+    }
 }
 
 pub mod misc {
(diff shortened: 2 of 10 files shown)


modules/axhal/src/platform/dummy/mod.rs
```diff
@@ -11,6 +11,17 @@ pub mod console {
     pub fn getchar() -> Option<u8> {
         unimplemented!()
     }
+
+    /// Writes bytes to the console from input u8 slice.
```
equation314, line 5: Do we still need `putchar` and `getchar`?

hky1999: > Do we still need `putchar` and `getchar`?

This is the interface that was left over from before. I'll keep it if it doesn't affect me.

modules/axhal/src/platform/riscv64_qemu_virt/console.rs
```diff
@@ -1,14 +1,28 @@
...
-pub fn getchar() -> Option<u8> {
-    #[allow(deprecated)]
-    match sbi_rt::legacy::console_getchar() as isize {
-        -1 => None,
-        c => Some(c as u8),
-    }
+/// Reads bytes from the console into the given mutable slice.
+/// Returns the number of bytes read.
+pub fn read_bytes(bytes: &mut [u8]) -> usize {
+    sbi_rt::console_read(sbi_rt::Physical::new(
```
equation314, line 41: Limit buffer size, like 256, the SBI seems to disable IRQs, we should not read too many bytes.

modules/axhal/src/platform/aarch64_bsta1000b/dw_apb_uart.rs
```diff
@@ -21,6 +21,28 @@ pub fn putchar(c: u8) {
     }
 }
 
+/// Write a slice of bytes to the console.
+pub fn write_bytes(bytes: &[u8]) {
```
equation314, line 30: Put `getchar` before this function.

hky1999: done, also make it un-public

modules/axhal/src/platform/riscv64_qemu_virt/console.rs
hky1999: should we use a warning here or simply use a assertion

equation314: Just use `bytes.len().min(MAX_READ_SIZE)`

hky1999: > Just use `bytes.len().min(MAX_READ_SIZE)`

ok, get it. But should we give user some feedback like a warning like bytes length exceeds the `MAX_READ_SIZE`

hky1999: since we return number of bytes to read in the return value.
i think no warning or other feedback should be given

luojia65 (approved):
LGTM. If possible, should we test the kernel with this patch on various SBI implementations?

hky1999:
> LGTM. If possible, should we test the kernel with this patch on various SBI implementations?

Thanks for your approval, as a matter of fact, it seems to failed to output anything on qemu 7.1.0

luojia65:
@hky1999: Does the built-in firmware in QEMU 7.1.0 support SBI 2.0 DBCN extension? If possible, we can use newer QEMU version, or other versions of any SBI firmware.

equation314 (approved):

hky1999:
> Does the built-in firmware in QEMU 7.1.0 support SBI 2.0 DBCN extension? 

@luojia65  I'm not familiar with that, how can I check that?

> If possible, we can use newer QEMU version, or other versions of any SBI firmware.

sure, qemu of version 9.0.0 and 8.2.0 all work fine.

luojia65:
> > Does the built-in firmware in QEMU 7.1.0 support SBI 2.0 DBCN extension?
> 
> @luojia65 I'm not familiar with that, how can I check that?

@hky1999 We can use sbi_rt::probe_extension feature to check if DBCN presents; if it returned 0, it means that this extension is available.

hky1999:
> > > Does the built-in firmware in QEMU 7.1.0 support SBI 2.0 DBCN extension?
> > 
> > 
> > @luojia65 I'm not familiar with that, how can I check that?
> 
> @hky1999 We can use sbi_rt::probe_extension feature to check if DBCN presents; if it returned 0, it means that this extension is available.

oh, i remembered, we use this api for the check of h-extension, ok we'll check it later

PeizhongQiu:
> > > Does the built-in firmware in QEMU 7.1.0 support SBI 2.0 DBCN extension?
> > 
> > 
> > @luojia65 I'm not familiar with that, how can I check that?
> 
> @hky1999 We can use sbi_rt::probe_extension feature to check if DBCN presents; if it returned 0, it means that this extension is available.

I checked it and found it returned 0

hky1999:
> > > > Does the built-in firmware in QEMU 7.1.0 support SBI 2.0 DBCN extension?
> > > 
> > > 
> > > @luojia65 I'm not familiar with that, how can I check that?
> > 
> > 
> > @hky1999 We can use sbi_rt::probe_extension feature to check if DBCN presents; if it returned 0, it means that this extension is available.
> 
> I checked it and found it returned 0

Do we need to consider compatibility design here, like rollback to legacy SBI when DBSN extension is not supported?
or it's just fine for us to stick on the latest SBI version here ：）
@luojia65 @equation314

PR #196, opened 2024-10-29
[fix] add build ci for other platforms, use read_bytes for terminate of x86_64-pc-oslab

hky1999:
fix the bug introduced by https://github.com/arceos-org/arceos/pull/195

getchar is required by `x86_64-pc-oslab` platform in `modules/axhal/src/platform/x86_pc/misc.rs`

```rust
/// Shutdown the whole system (in QEMU), including all CPUs.
///
/// See <https://wiki.osdev.org/Shutdown> for more information.
pub fn terminate() -> ! {
    info!("Shutting down...");

    #[cfg(platform = "x86_64-pc-oslab")]
    {
        axlog::ax_println!("System will reboot, press any key to continue ...");
        while super::console::getchar().is_none() {}
        axlog::ax_println!("Rebooting ...");
        unsafe { PortWriteOnly::new(0x64).write(0xfeu8) };
    }
}
```

also I re-introduced `build-for-other-platforms`, because this bug is found by the ci test from arceos-apps, I think it should be the arceos itself that is responsible for it

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,3 +81,63 @@ jobs:
       continue-on-error: ${{ matrix.rust-toolchain == 'nightly' }}
       run: make ARCH=${{ matrix.arch }} A=examples/httpserver-c
 
+  build-for-other-platforms:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        rust-toolchain: [nightly, nightly-2024-05-02]
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+      with:
+        toolchain: ${{ matrix.rust-toolchain }}
+        components: rust-src, llvm-tools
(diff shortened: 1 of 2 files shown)


equation314:
We should make the API list consistent across all platforms. So add `putchar` /`getchar` for all platforms, or use `console_read_bytes` for `x86_64-pc-oslab`

hky1999:
> We should make the API list consistent across all platforms. So add `putchar` /`getchar` for all platforms, or use `console_read_bytes` for `x86_64-pc-oslab`

I'll just use `read_bytes`, like this
```rust
    #[cfg(platform = "x86_64-pc-oslab")]
    {
        axlog::ax_println!("System will reboot, press any key to continue ...");
        let mut buffer = [0u8; 1];
        while super::console::read_bytes(&buffer) == 0 {}
        axlog::ax_println!("Rebooting ...");
        unsafe { PortWriteOnly::new(0x64).write(0xfeu8) };
    }
```

equation314 (dismissed):

equation314:
What's wrong with CI?

hky1999:
> What's wrong with CI?

It seems to be a github internal error, i do not know why

equation314 (approved):

PR #198, opened 2024-11-02
[fix] bug related with migrate_current and on_cpu flag polling

### LINKED ISSUE #197: Deadlock in `switch_to` Due to Premature Task Placement in Run Queue (SMP)
## Description

In `run_queue.rs`, the `switch_to` function in ArceOS currently implements a spin-loop to wait until the `on_cpu` flag for the next_task is set to `false`. 

This design is intended to ensure that the scheduling of next_task occurs atomically on the current CPU, preventing simultaneous access to the same task on different CPUs.

```rust
fn switch_to(&mut self, prev_task: CurrentTask, next_task: AxTaskRef) {
        // Make sure that IRQs are disabled by kernel guard or other means.
        #[cfg(all(not(test), feature = "irq"))] // Note: irq is faked under unit tests.
        assert!(
            !axhal::arch::irqs_enabled(),
            "IRQs must be disabled during scheduling"
        );
        debug!(
            "context switch: {} -> {}",
            prev_task.id_name(),
            next_task.id_name()
        );
        #[cfg(feature = "preempt")]
        next_task.set_preempt_pending(false);
        next_task.set_state(TaskState::Running);
        if prev_task.ptr_eq(&next_task) {
            return;
        }

        // Task must be scheduled atomically, wait for next task's scheduling process to complete.
        // If the owning (remote) CPU is still in the middle of schedule() with
        // this task (next task) as prev, wait until it's done referencing the task.
        //
        // Pairs with the `clear_prev_task_on_cpu()`.
        //
        // Note:
        // 1. This should be placed after the judgement of `TaskState::Blocked,`,
        //    because the task may have been woken up by other cores.
        // 2. This can be placed in the front of `switch_to()`
        #[cfg(feature = "smp")]
        while next_task.on_cpu() {
            // Wait for the task to finish its scheduling process.
            core::hint::spin_loop();
        }

        // Claim the task as running, we do this before switching to it
        // such that any running task will have this set.
        #[cfg(feature = "smp")]
        next_task.set_on_cpu(true);
        
        // ...
}
```

However, this design leads to a deadlock scenario under specific conditions in an SMP setup. 

If:
* Core 0 attempts to yield from `task 0` to `task 1`
* Core 1 simultaneously attempts to yield from `task 1` to `task 0`

Core 0 will be stuck waiting for `task 1` to fully yield Core 1, and Core 1 will be waiting for `task 0` to yield Core 0, resulting in a deadlock.

## Root Cause

The deadlock is caused by prematurely placing a task in the target core’s run queue (rq), allowing it to be immediately accessed and potentially popped by `switch_to` on that core. 

This access in `switch_to` leads to the aforementioned blocking spin-loop, as both cores await each other.

## Previous Solution

The polling should be placed at `unblock_task`, which would only insert the target task into the target CPU’s run queue once the `on_cpu` flag has been safely unset on the previous CPU. 

This will ensure the task is only placed into the run queue when it’s no longer active on another core.

## Complication with migrate_current

The reason why I place the polling in `switch_to` is that we need to support `migrate_current`, as I claimed in https://github.com/arceos-org/arceos/pull/183#discussion_r1799395422

Moving the polling to `unblock_task` introduces complexity for the `migrate_current` method. 

Since `migrate_current` is intended to place the current task in another CPU’s run queue, it would require waiting on the current task's `on_cpu` flag to be `false`—which is inherently impossible for the current task.

hky1999:
Fix Issue #197

### DIFF
--- a/modules/axtask/src/api.rs
+++ b/modules/axtask/src/api.rs
@@ -151,13 +151,12 @@ pub fn set_current_affinity(cpumask: AxCpuMask) -> bool {
     if cpumask.is_empty() {
         false
     } else {
-        let mut rq = current_run_queue::<NoPreemptIrqSave>();
         current().set_cpumask(cpumask);
         // After setting the affinity, we need to check if current cpu matches
         // the affinity. If not, we need to migrate the task to the correct CPU.
         #[cfg(feature = "smp")]
         if !cpumask.get(axhal::cpu::this_cpu_id()) {
-            rq.migrate_current();
+            current_run_queue::<NoPreemptIrqSave>().migrate_current();
         }
         true
     }
--- a/modules/axtask/src/api.rs
+++ b/modules/axtask/src/api.rs
@@ -155,6 +155,7 @@ pub fn set_current_affinity(cpumask: AxCpuMask) -> bool {
         current().set_cpumask(cpumask);
         // After setting the affinity, we need to check if current cpu matches
         // the affinity. If not, we need to migrate the task to the correct CPU.
+        #[cfg(feature = "smp")]
         if !cpumask.get(axhal::cpu::this_cpu_id()) {
             rq.migrate_current();
         }
--- a/modules/axtask/src/api.rs
+++ b/modules/axtask/src/api.rs
@@ -151,12 +151,24 @@ pub fn set_current_affinity(cpumask: AxCpuMask) -> bool {
     if cpumask.is_empty() {
         false
     } else {
-        current().set_cpumask(cpumask);
+        let curr = current().clone();
+
+        curr.set_cpumask(cpumask);
         // After setting the affinity, we need to check if current cpu matches
         // the affinity. If not, we need to migrate the task to the correct CPU.
         #[cfg(feature = "smp")]
         if !cpumask.get(axhal::cpu::this_cpu_id()) {
-            current_run_queue::<NoPreemptIrqSave>().migrate_current();
+            const MIGRATION_TASK_STACK_SIZE: usize = 4096;
+            // Spawn a new migration task for migrating.
+            let migration_task = TaskInner::new(
+                move || run_migrating(curr.clone()),
+                "migration-task".into(),
+                MIGRATION_TASK_STACK_SIZE,
+            )
+            .into_arc();
+
+            // Migrate the current task to the correct CPU using the migration task.
+            current_run_queue::<NoPreemptIrqSave>().migrate_current(migration_task);
         }
         true
     }
@@ -200,4 +212,12 @@ pub fn run_idle() -> ! {
         #[cfg(feature = "irq")]
         axhal::arch::wait_for_irqs();
     }
+}
+
+/// The task routine for migrating the current task to the correct CPU.
+///
+/// It calls `select_run_queue` to get the correct run queue for the task, and
+/// then puts the task to the run queue.
+fn run_migrating(migrated_task: AxTaskRef) {
+    select_run_queue::<NoPreemptIrqSave>(&migrated_task).put_prev_task(migrated_task);
 }
--- a/modules/axtask/src/api.rs
+++ b/modules/axtask/src/api.rs
@@ -161,7 +161,7 @@ pub fn set_current_affinity(cpumask: AxCpuMask) -> bool {
             const MIGRATION_TASK_STACK_SIZE: usize = 4096;
             // Spawn a new migration task for migrating.
             let migration_task = TaskInner::new(
-                move || migrate_entry(curr.clone()),
+                move || crate::run_queue::migrate_entry(curr),
                 "migration-task".into(),
                 MIGRATION_TASK_STACK_SIZE,
             )
@@ -214,12 +214,4 @@ pub fn run_idle() -> ! {
         #[cfg(feature = "irq")]
         axhal::arch::wait_for_irqs();
     }
-}
-
-/// The task routine for migrating the current task to the correct CPU.
-///
-/// It calls `select_run_queue` to get the correct run queue for the task, and
-/// then puts the task to the run queue.
-fn migrate_entry(migrated_task: AxTaskRef) {
-    select_run_queue::<NoPreemptIrqSave>(&migrated_task).put_prev_task(migrated_task);
 }
--- a/modules/axtask/src/api.rs
+++ b/modules/axtask/src/api.rs
@@ -161,7 +161,7 @@ pub fn set_current_affinity(cpumask: AxCpuMask) -> bool {
             const MIGRATION_TASK_STACK_SIZE: usize = 4096;
             // Spawn a new migration task for migrating.
             let migration_task = TaskInner::new(
-                move || run_migrating(curr.clone()),
+                move || migrate_entry(curr.clone()),
                 "migration-task".into(),
                 MIGRATION_TASK_STACK_SIZE,
             )
@@ -169,6 +169,8 @@ pub fn set_current_affinity(cpumask: AxCpuMask) -> bool {
 
             // Migrate the current task to the correct CPU using the migration task.
             current_run_queue::<NoPreemptIrqSave>().migrate_current(migration_task);
+
+            assert!(cpumask.get(axhal::cpu::this_cpu_id()), "Migration failed");
         }
         true
     }
@@ -218,6 +220,6 @@ pub fn run_idle() -> ! {
 ///
 /// It calls `select_run_queue` to get the correct run queue for the task, and
 /// then puts the task to the run queue.
-fn run_migrating(migrated_task: AxTaskRef) {
+fn migrate_entry(migrated_task: AxTaskRef) {
     select_run_queue::<NoPreemptIrqSave>(&migrated_task).put_prev_task(migrated_task);
 }
--- a/modules/axtask/src/run_queue.rs
+++ b/modules/axtask/src/run_queue.rs
@@ -2,7 +2,7 @@ use alloc::collections::VecDeque;
 use alloc::sync::Arc;
 use core::mem::MaybeUninit;
 
-use kernel_guard::{BaseGuard, NoOp};
+use kernel_guard::BaseGuard;
 use kspin::SpinRaw;
 use lazyinit::LazyInit;
 use scheduler::BaseScheduler;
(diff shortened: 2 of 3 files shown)


modules/axtask/src/task.rs
```diff
@@ -58,7 +58,7 @@ pub struct TaskInner {
     on_cpu: AtomicBool,
     /// A weak reference to the previous task running on this CPU.
     #[cfg(feature = "smp")]
-    prev_task: UnsafeCell<Weak<AxTask>>,
+    prev_task: UnsafeCell<Option<AxTaskRef>>,
```
hky1999: @equation314 please mind here, I use `Arc` instead of `Weak`, because we need a place to store the reference to "migrated task" which is not put into the run queue
details are explained in https://github.com/arceos-org/arceos/issues/197#issuecomment-2452913756

modules/axtask/src/run_queue.rs
```diff
@@ -247,6 +252,18 @@ impl<'a, G: BaseGuard> AxRunQueueRef<'a, G> {
...
     }
+
+    /// Migrate the task to this run queue.
+    ///
+    /// Panics if the task is not in [`TaskState::Migrating`].
+    pub fn migrate_task(&mut self, task: AxTaskRef) {
+        let task_id_name = task.id_name();
+        let cpu_id = self.inner.cpu_id;
+        debug!("task migrate: {} to run_queue {}", task_id_name, cpu_id);
+        assert!(self
```
equation314: Put the function call outside `assert`

modules/axtask/src/run_queue.rs
```diff
@@ -282,9 +299,7 @@ impl<'a, G: BaseGuard> CurrentRunQueueRef<'a, G> {
         trace!("task migrate: {}", curr.id_name());
         assert!(curr.is_running());
 
-        select_run_queue::<NoOp>(curr.as_task_ref())
-            .inner
-            .put_task_with_state(curr.clone(), TaskState::Running, false);
+        curr.set_state(TaskState::Migrating);
 
         self.inner.resched();
```
equation314: Maybe we can introduce a new function `migrate_resched` to reduce the interference of the regular rescheduling.

modules/axtask/src/task.rs
equation314: I think we should still use the `Weak` pointer to save `prev_task`, but keep an additional per-CPU variable `migrating_task` to avoid affecting the reference count of normal tasks.

BTW, I found that `prev_task` can also be a per-CPU variable instead of a per-task variable to simplify the `TaskInner` structure.

hky1999: Really???

yep it would be more convenient, I just thought that it's ugly to have so many global percore variables...

modules/axtask/src/run_queue.rs
equation314: We can just select run queue and put the previous task after `resched`.

hky1999: you mean, let "next task" to put previous task into run queue?

modules/axtask/src/task.rs
hky1999: > I think we should still use the `Weak` pointer to save `prev_task`, but keep an additional per-CPU variable `migrating_task` to avoid affecting the reference count of normal tasks.

The `Weak` pointer for `prev_task` is just used to finished the scheduling process of prev task by setting its `on_cpu` flag as false.

> BTW, I found that prev_task can also be a per-CPU variable instead of a per-task variable to simplify the TaskInner structure.

if `prev_task` is stored in a per-CPU variable, yes, we can simply put all operations related with prev task after `resched`, just put them in `finish_prev_task_switch`, including selecting a run queue for prev task due to normal yielding

modules/axtask/src/run_queue.rs
equation314: Yes, move `finish_prev_task_switch ` after `self.inner.resched()`, and leave `clear_prev_task_on_cpu` in `switch_to`.

hky1999: > Yes, move `finish_prev_task_switch ` after `self.inner.resched()`, and leave `clear_prev_task_on_cpu` in `switch_to`.

`finish_prev_task_switch` can not be placed after `self.inner.resched()`, because it should be the next task that is responsible for finishing the scheduling process of previous task.

I still combined `clear_prev_task_on_cpu` into `finish_prev_task_switch`, and put it in `switch_to`

modules/axtask/src/api.rs
```diff
@@ -155,6 +155,7 @@ pub fn set_current_affinity(cpumask: AxCpuMask) -> bool {
         current().set_cpumask(cpumask);
         // After setting the affinity, we need to check if current cpu matches
         // the affinity. If not, we need to migrate the task to the correct CPU.
+        #[cfg(feature = "smp")]
         if !cpumask.get(axhal::cpu::this_cpu_id()) {
```
hky1999, line 160: what should we do if "smp" feature is not enabled??? @equation314

equation314: That's ok, set `cpumask` only and do not migrate.

modules/axtask/src/run_queue.rs
```diff
@@ -14,8 +17,24 @@ use crate::wait_queue::WaitQueueGuard;
 use crate::{AxCpuMask, AxTaskRef, Scheduler, TaskInner, WaitQueue};
 
 macro_rules! percpu_static {
-    ($($name:ident: $ty:ty = $init:expr),* $(,)?) => {
+    ($(
+        $(#[$comment:meta])*
+        #[cfg($($cfg:tt)*)]
```
equation314: Just use `$(#[$comment:meta])*`.

modules/axtask/src/api.rs
hky1999: ok, and this PR is ready for review

modules/axtask/src/run_queue.rs
```diff
@@ -567,6 +610,21 @@ fn gc_entry() {
     }
 }
 
+#[cfg(feature = "smp")]
+pub(crate) unsafe fn finish_task_switch() {
```
equation314: This can be divided into two, one is still `clear_prev_task_on_cpu`, and one is selecting run queue for the migrated task after `resched` in `migrate_current`, so that we don't need to check `MIGRATING_TASK` during normal task switching.

hky1999: > This can be divided into two, one is still `clear_prev_task_on_cpu`, and one is selecting run queue for the migrated task after `resched` in `migrate_current`, so that we don't need to check `MIGRATING_TASK` during normal task switching.

I understand your intension. 

BUT, the question is, when a task calls migrate_current, it's the next task that is reponsible to check whether `MIGRATING_TASK` is not `None`, we can not just put something like selecting run queue for the migrated task after `resched` in `migrate_current`

modules/axtask/src/run_queue.rs
hky1999: updated

modules/axtask/src/run_queue.rs
equation314: Why? After `resched`, current task is the next task, just like `clear_prev_task_on_cpu` on `switch_to`.

hky1999: Yep, but it will run `switch_to` belonging to next task (or `task_entry` belongs to next task), which will never return to the `resched` in `migrate_current`.

modules/axtask/src/task.rs
```diff
@@ -410,50 +403,19 @@ impl TaskInner {
     /// It is used to protect the task from being moved to a different run queue
     /// while it has not finished its scheduling process.
     /// The `on_cpu field is set to `true` when the task is preparing to run on a CPU,
-    /// and it is set to `false` when the task has finished its scheduling process in `clear_prev_task_on_cpu()`.
+    /// and it is set to `false` when the task has finished its scheduling process in `finish_task_switch()`.
```
equation314: Fix comments

modules/axtask/src/run_queue.rs
```diff
@@ -521,9 +554,11 @@ impl AxRunQueue {
             let prev_ctx_ptr = prev_task.ctx_mut_ptr();
             let next_ctx_ptr = next_task.ctx_mut_ptr();
 
-            // Store the weak pointer of **prev_task** in **next_task**'s struct.
+            // Store the weak pointer of **prev_task** in percpu variable `PREV_TASK`.
             #[cfg(feature = "smp")]
-            next_task.set_prev_task(prev_task.as_task_ref());
+            {
```
equation314, line 553: Remove `{}`

modules/axtask/src/run_queue.rs
```diff
@@ -275,18 +290,33 @@ impl<'a, G: BaseGuard> CurrentRunQueueRef<'a, G> {
...
+                curr.set_state(TaskState::Ready);
+                select_run_queue::<kernel_guard::NoPreemptIrqSave>(&curr).add_task(curr);
+            },
+            "migration-task".into(),
+            MIGRATION_TASK_STACK_SIZE,
+        )
+        .into_arc();
+
+        // Call `switch_to` to reschedule to the migration task that performs the migration directly.
+        self.inner.switch_to(crate::current(), migration);
```
equation314: why not use `curr`?

modules/axtask/src/run_queue.rs
```diff
@@ -275,18 +290,34 @@ impl<'a, G: BaseGuard> CurrentRunQueueRef<'a, G> {
...
-            .put_task_with_state(curr.clone(), TaskState::Running, false);
-
-        self.inner.resched();
+        const MIGRATION_TASK_STACK_SIZE: usize = 4096;
+
+        // Do not put current task to the run queue,
+        // instead, spawn a new task for migrating.
+        let migration = TaskInner::new(
+            move || {
+                curr.set_state(TaskState::Ready);
```
equation314: Why not update `curr`'s state outside the migration task?

modules/axtask/src/run_queue.rs
hky1999: `PREV_TASK` is a global percpu static variable decorated by percpu macros, remove this `{}` may cause some grammer problems?

modules/axtask/src/run_queue.rs
hky1999: the first argument of `switch_to` is `CurrentTask`, which does not implement `Copy`

equation314 (approved):

PR #205, opened 2024-11-27
[feat] add nvme support

ZR233:
1. add aarch64 dcache clean and invalid funcs.
2. add nvme driver.

### DIFF
--- a/modules/axdriver/src/bus/pci.rs
+++ b/modules/axdriver/src/bus/pci.rs
@@ -1,8 +1,11 @@
+use core::ptr::NonNull;
+
 use crate::{prelude::*, AllDevices};
 use axdriver_pci::{
     BarInfo, Cam, Command, DeviceFunction, HeaderType, MemoryBarType, PciRangeAllocator, PciRoot,
 };
 use axhal::mem::phys_to_virt;
+use pcie::{Header, RootComplexGeneric};
 
 const PCI_BAR_NUM: u8 = 6;
 
@@ -84,6 +87,33 @@ fn config_pci_device(
 
 impl AllDevices {
     pub(crate) fn probe_bus_devices(&mut self) {
+        self.probe_pci_devices();
+        self.probe_pcie_devices();
+    }
+
+    fn probe_pcie_devices(&mut self) {
+        let base_vaddr = phys_to_virt(axconfig::PCI_ECAM_BASE.into());
+        let mut root = RootComplexGeneric::new(NonNull::new(base_vaddr.as_mut_ptr()).unwrap());
+
+        for elem in root.enumerate_keep_bar(None) {
+            if let Header::Endpoint(ep) = elem.header {
+                for_each_drivers!(type Driver, {
+                    if let Some(dev) = Driver::probe_pcie(elem.root, &ep) {
+                        info!(
+                            "registered a new {:?} device at {:?}: {:?}",
+                            dev.device_type(),
+                            ep.address,
+                            dev.device_name(),
+                        );
+                        self.add_device(dev);
+                        continue; // skip to the next device
+                    }
+                })
+            }
+        }
+    }
+
+    fn probe_pci_devices(&mut self) {
         let base_vaddr = phys_to_virt(axconfig::PCI_ECAM_BASE.into());
         let mut root = unsafe { PciRoot::new(base_vaddr.as_mut_ptr(), Cam::Ecam) };
 
--- a/modules/axdriver/src/lib.rs
+++ b/modules/axdriver/src/lib.rs
@@ -78,6 +78,9 @@ mod virtio;
 #[cfg(feature = "ixgbe")]
 mod ixgbe;
 
+#[cfg(feature = "nvme")]
+mod nvme;
+
 pub mod prelude;
 
 #[allow(unused_imports)]
--- a/modules/axhal/src/arch/riscv/cache.rs
+++ b/modules/axhal/src/arch/riscv/cache.rs
@@ -1,0 +1,13 @@
+#![allow(unused)]
+
+use core::ptr::NonNull;
+
+/// Invalidate data cache
+pub fn dcache_invalidate_range(_addr: NonNull<u8>, _size: usize) {
+    unimplemented!();
+}
+
+/// Flush data cache
+pub fn dcache_flush_range(_addr: NonNull<u8>, _size: usize) {
+    unimplemented!();
+}
--- a/modules/axhal/src/arch/x86_64/cache.rs
+++ b/modules/axhal/src/arch/x86_64/cache.rs
@@ -1,0 +1,13 @@
+#![allow(unused)]
+
+use core::ptr::NonNull;
+
+/// Invalidate data cache
+pub fn dcache_invalidate_range(_addr: NonNull<u8>, _size: usize) {
+    unimplemented!();
+}
+
+/// Flush data cache
+pub fn dcache_flush_range(_addr: NonNull<u8>, _size: usize) {
+    unimplemented!();
+}
--- a/modules/axhal/src/dma.rs
+++ b/modules/axhal/src/dma.rs
@@ -1,0 +1,27 @@
+use dma_api::Impl;
+
+use crate::{
+    arch::cache::{dcache_flush_range, dcache_invalidate_range},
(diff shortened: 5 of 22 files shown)


modules/axdriver/src/bus/pci.rs
```diff
@@ -1,8 +1,12 @@
+use core::ptr::NonNull;
+
 use crate::{prelude::*, AllDevices};
+use alloc::vec::Vec;
```
hky1999: move this dep to the beginning of this file, together with `use core::*`

modules/axhal/src/arch/riscv/cache.rs
```diff
@@ -0,0 +1,13 @@
+#![allow(unused)]
+
+use core::ptr::NonNull;
+
+/// Invalidate data cache
+pub fn dcache_invalidate_range(_addr: NonNull<u8>, _size: usize) {
+    //TODO
```
hky1999: unimplemented!

modules/axhal/src/arch/x86_64/cache.rs
```diff
@@ -0,0 +1,13 @@
+#![allow(unused)]
+
+use core::ptr::NonNull;
+
+/// Invalidate data cache
+pub fn dcache_invalidate_range(_addr: NonNull<u8>, _size: usize) {
+    //TODO
```
hky1999: unimplemented

modules/axhal/src/lib.rs
```diff
@@ -57,6 +57,9 @@ pub mod irq;
 #[cfg(feature = "paging")]
 pub mod paging;
 
+#[cfg(feature = "alloc")]
+mod dma;
```
hky1999, line 61: this is extremely strange, why the dma mod is decorated under `#[cfg(feature = "alloc")]`

modules/axhal/src/dma.rs
```diff
@@ -0,0 +1,26 @@
+use dma_api::Impl;
```
hky1999, line 1: I think the `Impl` trait exposed by `dma-api` crate is strange.

I believe that you want to use this trait for the translation from `IOVA` to actual `PA`

hky1999: maybe you can refer to the DMA addr related trait defined in `virtio-drivers`? 

https://github.com/rcore-os/virtio-drivers/blob/master/src/hal.rs

scripts/make/qemu.mk
```diff
@@ -26,9 +26,14 @@ qemu_args-aarch64 := \
 
 qemu_args-y := -m 128M -smp $(SMP) $(qemu_args-$(ARCH))
 
-qemu_args-$(BLK) += \
-  -device virtio-blk-$(vdev-suffix),drive=disk0 \
-  -drive id=disk0,if=none,format=raw,file=$(DISK_IMG)
+ifeq ($(findstring driver-nvme,$(FEATURES)), driver-nvme)
```
hky1999: I am not sure if this can be configured through env var `$(FEATURES)`, if the user only enable this feature in `Cargo.toml`, it may not exist in `$(FEATURES)` @equation314

modules/axdriver/src/lib.rs
```diff
@@ -61,7 +61,7 @@
 #[macro_use]
 extern crate log;
 
-#[cfg(feature = "dyn")]
+#[cfg(any(feature = "dyn", feature = "bus-pci"))]
```
hky1999: this is awkward, extremely awkward

modules/axdriver/src/bus/pci.rs
```diff
@@ -84,6 +88,37 @@ fn config_pci_device(
...
     pub(crate) fn probe_bus_devices(&mut self) {
+        self.probe_pci_devices();
+        self.probe_pcie_devices();
+    }
+
+    fn probe_pcie_devices(&mut self) {
+        let base_vaddr = phys_to_virt(axconfig::PCI_ECAM_BASE.into());
+        let mut root = RootComplexGeneric::new(NonNull::new(base_vaddr.as_mut_ptr()).unwrap());
+
+        let headers = root.enumerate_no_modify(None).collect::<Vec<_>>();
```
hky1999: if you can convert it into slice.

deps related with `alloc` can be removed

modules/axhal/src/arch/riscv/cache.rs
ZR233: It can work with qemu, because virt machine cache is synced, but will not work on real machine. I am not familiar with riscv, so need someone else to impl this.

modules/axhal/src/arch/x86_64/cache.rs
ZR233: Same as riscv.

modules/axhal/src/arch/riscv/cache.rs
hky1999: > It can work with qemu, because virt machine cache is synced, but will not work on real machine. I am not familiar with riscv, so need someone else to impl this.

yep, I mean you should use `unimplemented()` macros explicitly, in case if someone want to use this API but failed to get the expected result.

modules/axhal/src/lib.rs
ZR233: Dma-api use normal alloc api to alloc Vec.

modules/axhal/src/dma.rs
ZR233: This crate is not just about converting virtual addresses into physical addresses，but also got dma direction usage by `map` func, when read from data, crate will invlid cache, when write, it will flush back from cache to memory, so that we can use normal type memory for dma, and direct use upstream &[u8] without copy.

modules/axhal/src/lib.rs
hky1999: > Dma-api use normal alloc api to alloc Vec.

I just do not get it, generally, the size of dma region is known and fixed, why a `Vec` is needed since it's usually related to dynamically arrays with unknown size.

modules/axhal/src/dma.rs
ZR233: When use `&mut [u8]` from os, `map` to dma slice type, and when give back &[u8] to os, `umap` to normal memory. If use like `virtio-drivers`, we need to copy `&[u8]` to no cache type memory for dma and copy back to `&[u8]` for send to os.

modules/axdriver/src/bus/pci.rs
ZR233: Need mutable borrow `root` later, so need two step.

modules/axhal/src/arch/riscv/cache.rs
ZR233: If use `unimplemented!()`, x86 qemu will not work, but it can work.

modules/axhal/src/lib.rs
ZR233: dma use global allocator.

hky1999: maybe we can delete this now?

modules/axhal/src/dma.rs
hky1999: Sorry, I need to learn how `dma_api` works before offering any more valuable suggestions.

kind of busy lately, so this may take a little bit longer

modules/axhal/src/dma.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn flush(addr: core::ptr::NonNull<u8>, size: usize) {
+        dcache_clean_range(addr, size);
+    }
+
+    fn invalidate(addr: core::ptr::NonNull<u8>, size: usize) {
+        dcache_invalidate_range(addr, size);
+    }
+}
+
+dma_api::set_impl!(DmaImpl);
```
hky1999, line 27: BTW, this is kind of strange, too.

ZR233: It a good way for singleton trait. It is widely used in embedded-hal group, like [critical-section](https://crates.io/crates/critical-section)

scripts/make/qemu.mk
equation314: @ZR233 You could add a new make argument `BLK_DEV` like `NET_DEV`.

PR #210, opened 2024-12-09
[feat] Add aarch64 and riscv dcache manage

ZR233:
Add dcache invalidate flush api.

### DIFF
--- a/modules/axhal/src/arch/aarch64/cache.rs
+++ b/modules/axhal/src/arch/aarch64/cache.rs
@@ -1,0 +1,106 @@
+use core::arch::asm;
+
+use aarch64_cpu::{asm::barrier::*, registers::*};
+use memory_addr::VirtAddr;
+
+use crate::arch::CacheOp;
+
+#[inline(always)]
+fn cache_line_size() -> usize {
+    unsafe {
+        let mut ctr_el0: u64;
+        asm!("mrs {}, ctr_el0", out(reg) ctr_el0);
+        let log2_cache_line_size = ((ctr_el0 >> 16) & 0xF) as usize;
+        // Calculate the cache line size
+        4 << log2_cache_line_size
+    }
+}
+
+/// Performs a cache operation on all memory.
+pub fn dcache_all(op: CacheOp) {
+    let clidr = CLIDR_EL1.get();
+
+    for level in 0..8 {
+        let ty = (clidr >> (level * 3)) & 0b111;
+        if ty == 0 {
+            return;
+        }
+
+        dcache_level(op, level);
+    }
+    dsb(SY);
+    isb(SY);
+}
+
+/// Performs a cache operation on a range of memory.
+#[inline]
+pub fn dcache_range(op: CacheOp, addr: VirtAddr, size: usize) {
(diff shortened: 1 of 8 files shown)


modules/axhal/src/arch/aarch64/cache.rs
equation314, line 1: Too many magic assembly code, try to rewrite it as much Rust code as possible.

equation314:
Currently `flush_dcache_line` and `flush_icache_all` have been implemented in axhal/src/arch/aarch64/mod.rs.

Could you abstract a unified set of icache and dcache management APIs? And clarify the difference between flush and invalidate.

For example, these APIs might be like:

* flush_dcache_line(op, addr)
* flush_dcache_range(op, start, end)
* flush_dcache_all(op)
* flush_icache_line(op, addr)
* flush_icache_range(op, start, end)
* flush_icache_all(op)


Where `op` is one of Clean/Invalidate/CleanAndInvalidate.

modules/axhal/src/arch/aarch64/cache.rs
```diff
@@ -0,0 +1,136 @@
...
+    unsafe {
+        let mut ctr_el0: u64;
+        asm!("mrs {}, ctr_el0", out(reg) ctr_el0);
+        let log2_cache_line_size = ((ctr_el0 >> 16) & 0xF) as usize;
+        // Calculate the cache line size
+        4 << log2_cache_line_size
+    }
+}
+
+struct DCacheIter {
```
equation314: Introducing an `Iterator` is over-design, a single `while` loop is simpler and cleaner.

modules/axhal/src/arch/aarch64/cache.rs
```diff
@@ -0,0 +1,136 @@
...
+    _dcache_line(op, addr);
+    unsafe {
+        asm!("dsb sy; isb");
+    }
+}
+
+/// Performs a cache operation on a cache level.
+/// https://developer.arm.com/documentation/ddi0601/2024-12/AArch64-Instructions/DC-CISW--Data-or-unified-Cache-line-Clean-and-Invalidate-by-Set-Way
+#[inline]
+fn dcache_level(op: CacheOp, level: u64) {
```
equation314, line 76: Flushing the entire D-cache is hardly ever used, so we will not implement it to avoid this cumbersome function.

modules/axhal/src/arch/aarch64/cache.rs
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+}
+
+/// Performs a cache operation on a cache line.
+#[inline]
+pub fn dcache_line(op: CacheOp, addr: VirtAddr) {
+    _dcache_line(op, addr);
+    unsafe {
+        asm!("dsb sy; isb");
```
equation314: use `aarch64_cpu::barrier::{dsb, isb}`

modules/axhal/src/arch/aarch64/cache.rs
ZR233: before enable mmu we need to flush all cache,  or will cause error. tested on phytium.

PR #217, opened 2025-01-06
[dockerfile] prepare toolchain

ZR233:
Therefore, you don't need to re-download the toolchain each time you create a container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,10 @@ RUN apt-get update \
 
 RUN cargo install cargo-binutils axconfig-gen
 
+COPY rust-toolchain.toml /rust-toolchain.toml
+
+RUN rustc --version
+
 RUN wget https://musl.cc/aarch64-linux-musl-cross.tgz \
     && wget https://musl.cc/riscv64-linux-musl-cross.tgz \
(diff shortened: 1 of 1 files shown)


equation314:
It may not be a good practice. We need to provide an environment as clean as possible in docker with no dependencies installed to have a chance to find some dependency related issues.

Dockerfile
```diff
@@ -6,6 +6,16 @@ RUN apt-get update \
 
 RUN cargo install cargo-binutils axconfig-gen
 
+RUN cargo init /rust-test
+
+WORKDIR /rust-test
+
+COPY rust-toolchain.toml /rust-test/rust-toolchain.toml
+
+RUN cargo build
```
equation314: `cargo build` is not needed, you can just run `rustc --version` in the `/` directory.

Dockerfile
```diff
@@ -6,6 +6,10 @@ RUN apt-get update \
 
 RUN cargo install cargo-binutils axconfig-gen
 
+COPY rust-toolchain.toml /rust-toolchain.toml
```
equation314, line 9: `rust-toolchain.toml` is already exsited in `/`.

equation314 (approved):

PR #223, opened 2025-02-06
Fix app test on httpclient

equation314:
See https://github.com/arceos-org/arceos-apps/pull/6.

Also updated dependencies.


hky1999 (approved):
LGTM.
btw, I am kind of forgetting when did we bump percpu version to 0.2...

PR #226, opened 2025-03-27
[feat] aarch64 use dtb to find debug uart

ZR233:
1. find uart by dtb.
2. aux、8250、pl011 support all in one.
3. aarch64-bsta1000b should work with this, but we don't have hardwire to test, so did not edit.


modules/axhal/src/platform/aarch64_common/debug.rs
```diff
@@ -0,0 +1,67 @@
+use core::ptr::NonNull;
+
+use any_uart::*;
```
hky1999, line 3: you just totally ignore dtb parsing in arceos, but just forward dtb address to the `any_uart`.

It seems quite strange, if we want to parse more device info from device tree, do we need to parse the device info in other crates, again and again?

hky1999:
@equation314 Is it a proper timer for us to introduce DTB into ArceOS.

modules/axhal/src/platform/aarch64_common/debug.rs
ZR233: This crate is only used for quick debugging. Other bare metal testing environments can directly obtain debug functionality by providing only the dtb address. The actual device tree traversal should be done during the initialization process later.

ZR233: Full func uart device should be inited in driver enum process.

equation314:
We should not introduce DTB in ArceOS backbone.  DTB parsing can be used for some specific platforms (e.g., aarch64-phytium-pi), not all. Platforms are allowed to not use DTB to keep the code base minimal.

It's also a bit strange to include DTB parsing in the `any_uart` crate, which should only be responsible for UART device operations.

PR #228, opened 2025-05-22
Upgrade toolchain to nightly-2025-05-20

equation314:
* Change `#[naked]` to `#[unsafe(naked)]` to meet the latest toolchain requirements.
* Fix bindgen generation error due to invalid `softfloat` targets used by clang.
* Fix kernel hangs caused by LoongArch64 LSX (SIMD) instructions.

### DIFF
--- a/modules/axdriver/src/lib.rs
+++ b/modules/axdriver/src/lib.rs
@@ -40,12 +40,12 @@
 //! - `dyn`: use the dynamic device model (see above).
 //! - `bus-mmio`: use device tree to probe all MMIO devices.
 //! - `bus-pci`: use PCI bus to probe all PCI devices. This feature is
-//!    enabeld by default.
+//!   enabled by default.
 //! - `virtio`: use VirtIO devices. This is enabled if any of `virtio-blk`,
 //!   `virtio-net` or `virtio-gpu` is enabled.
 //! - `net`: use network devices. This is enabled if any feature of network
-//!    devices is selected. If this feature is enabled without any network device
-//!    features, a dummy struct is used for [`AxNetDevice`].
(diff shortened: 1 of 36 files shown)


modules/axdriver/src/lib.rs
```diff
@@ -40,12 +40,12 @@
 //! - `dyn`: use the dynamic device model (see above).
 //! - `bus-mmio`: use device tree to probe all MMIO devices.
 //! - `bus-pci`: use PCI bus to probe all PCI devices. This feature is
-//!    enabeld by default.
+//!   enabeld by default.
```
hky1999: enabled

hky1999 (dismissed):

Azure-stars:
Should we solve the https://github.com/arceos-org/arceos/issues/229 in this PR?

Azure-stars:
And maybe we need to clean some cached files. See https://github.com/arceos-org/arceos/actions/caches.It is about to reach the upper limit of the cache capacity, causing the cache in CI to fail to save and report an warning(See annotations in https://github.com/arceos-org/arceos/actions/runs/15188591377).

Azure-stars (dismissed):

equation314:
> And maybe we need to clean some cached files. See https://github.com/arceos-org/arceos/actions/caches.It is about to reach the upper limit of the cache capacity, causing the cache in CI to fail to save and report an warning(See annotations in https://github.com/arceos-org/arceos/actions/runs/15188591377).

There a lot of unused caches starts with `v0-rust-*`, maybe we need to use [Swatinem/rust-cache@v2](https://github.com/Swatinem/rust-cache/tree/v2/) correctly.

Azure-stars (approved):

PR #235, opened 2025-06-03
Use setup-musl & setup-qemu actions from Marketplace

equation314:
1. Make CI happy: use [setup-musl](https://github.com/marketplace/actions/download-and-build-qemu) & [setup-musl](https://github.com/marketplace/actions/setup-the-musl-toolchain) actions from Marketplace.
2. Use `shared-key` to further reduce the cache size of GitHub Action.


Azure-stars (approved):

.github/workflows/build.yml
```diff
@@ -90,6 +92,8 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         rust-toolchain: [nightly, nightly-2025-05-20]
+    env:
+      RUSTUP_TOOLCHAIN: ${{ matrix.rust-toolchain }}
```
Azure-stars, line 96: Why we need to set this environment variable? For `shared-key`?

equation314: This is a previous bug. Without force toolchain setting by `env`, cargo will always use the nightly-2025-05-20 toolchain in all cases.

PR #237, opened 2025-06-06
Record the boot_stack in `kstack` field in `TaskInner` while creating init task.

zflcs:
- [hal]: Add `read_sp` function to get the stack pointer, `current_boot_stack` to get boot stack base. 
- [task]: Add `is_init` field and `new_init` function for `TaskStack`, and record the boot_stack in `kstack` field in `TaskInner` while creating init task.


Azure-stars:
In what scenarios will the modifications you submit be used?

zflcs:
> In what scenarios will the modifications you submit be used?

I'm trying to add support about Rust coroutine and the `kstack` field is used to help identify whether a task is a coroutine or thread. If the `kstack` field of init task is None, the task-switching function will treat the init task as a coroutine.

PR #244, opened 2025-06-10
fix: make `AxNamespace` `Send + Sync`

elebirds:

For further information, please refer to: [fix: make AxNamespace Send + Sync #30](https://github.com/oscomp/arceos/pull/30)

## Description
This PR changes the base field in AxNamespace from a raw pointer to a address, to make this type Send and Sync. This is necessary for sharing namespaces, otherwise we would have to trick the compiler.

It's the prerequisite requirements of https://github.com/oscomp/starry-next/pull/34

## Additional Notes

### DIFF
--- a/modules/axns/src/lib.rs
+++ b/modules/axns/src/lib.rs
@@ -49,6 +49,9 @@ pub struct AxNamespace {
     alloc: bool,
 }
 
+unsafe impl Send for AxNamespace {}
+unsafe impl Sync for AxNamespace {}
+
 impl AxNamespace {
     /// Returns the base address of the namespace, which points to the start of
     /// all resources.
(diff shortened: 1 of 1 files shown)


modules/axns/src/lib.rs
```diff
@@ -45,15 +45,15 @@ unsafe extern "C" {
 ///   consistent. Each namespace has its own resources, which may be unique or
 ///   shared between threads by the [`Arc`] wrapper.
 pub struct AxNamespace {
-    base: *mut u8,
+    base: usize,
```
equation314: I suggest keeping the original pointer type and manually adding `unsafe impl Send for AxNamespace`.

elebirds: Fine, in my view both solutions are acceptables —— there's no obvious difference between them. I'll change it back to the original pointer type and manually adding `Send + Sync` trait.

Azure-stars: To use `*mut u8` type for the base pointer is more clear for me?

equation314 (approved):

PR #245, opened 2025-06-10
feat: build without axstd

elebirds:

For further information, please refer to: [feat: build without axstd #21](https://github.com/oscomp/arceos/pull/21)

## Description
This PR fixes this by introducing a `NO_AXSTD` option (similar to `no_std` in Rust) to use `axfeat/` as `ax_feat_prefix` and `AX_LIB` to control `lib_feat_prefix` (note that this is separate from `NO_AXSTD` as apps might want to use arceos_posix_api as `AX_LIB`).

## Additional Notes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,9 +75,13 @@ endif
 
 ifneq ($(wildcard $(APP)/Cargo.toml),)
   APP_TYPE := rust
+  AX_LIB ?= axstd
 else
   APP_TYPE := c
+  AX_LIB ?= axlibc
 endif
+
+NO_AXSTD ?= n
 
 # Feature parsing
 include scripts/make/features.mk
--- a/scripts/make/features.mk
+++ b/scripts/make/features.mk
@@ -13,14 +13,17 @@
 
 ifeq ($(APP_TYPE),c)
   ax_feat_prefix := axfeat/
-  lib_feat_prefix := axlibc/
   lib_features := fp_simd irq alloc multitask fs net fd pipe select epoll
 else
-  # TODO: it's better to use `axfeat/` as `ax_feat_prefix`, but all apps need to have `axfeat` as a dependency
-  ax_feat_prefix := axstd/
-  lib_feat_prefix := axstd/
+  ifeq ($(NO_AXSTD),y)
+    ax_feat_prefix := axfeat/
+  else
+    ax_feat_prefix := axstd/
+  endif
   lib_features :=
 endif
+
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -75,10 +75,14 @@ endif
 
 ifneq ($(wildcard $(APP)/Cargo.toml),)
   APP_TYPE := rust
+  AX_LIB ?= axstd
```
Azure-stars, line 78: Can we not use `AX_LIB` and keep `lib_feat_prefix` in the original place?

scripts/make/features.mk
```diff
@@ -13,15 +13,18 @@
...
 ifeq ($(APP_TYPE),c)
   ax_feat_prefix := axfeat/
-  lib_feat_prefix := axlibc/
   lib_features := fp_simd irq alloc multitask fs net fd pipe select epoll
 else
-  # TODO: it's better to use `axfeat/` as `ax_feat_prefix`, but all apps need to have `axfeat` as a dependency
-  ax_feat_prefix := axstd/
-  lib_feat_prefix := axstd/
+  ifeq ($(NO_AXSTD),y)
+    ax_feat_prefix := axfeat/
```
Azure-stars, line 19: When would we not use axstd build? For rust std and core lib, the latter can support building on bare metal, while the introduction of the former requires specific system support. But for `axstd`, there seems to be no such problem?

Makefile
elebirds: yep, I think it could keep it like the past

elebirds: for now, we could keep it like the past.
but you can see the original PR, it mentioned a possibility that apps might want to use arceos_posix_api as AX_LIB, in this situation, keeping `AX_LIB` seems like a better choice.

scripts/make/features.mk
elebirds: same as above

AsakuraMizu:
As the original author, I would like to suggest an alternative.

First I should mention my considerations at that time. Due to the defects in the build toolchain, for Rust programs, `FEATURES` are mapped to `axstd/xxx`, which prevents either removing the `axstd` dependency or `arceos_posix_api`.

But I later realized that it is not "authentic" to enable feature flags in this way for Rust programs. Developers should always specify the feature list of dependencies only in `Cargo.toml`.

So my suggestion would be to delete this part, use the native Rust to process options (including smp, log level and others). We only need to care about the features required by a C project.

PR #246, opened 2025-06-10
feat: introduced the `check_region_access` method in the `AddrSpace`

elebirds:

## Description
`check_region_access` verifies if a given memory region is fully mapped and has the required access permissions (`MappingFlags`). It iterates through memory areas, checking for overlaps and validating permission flags.

## Additional Notes

### DIFF
--- a/modules/axmm/src/aspace.rs
+++ b/modules/axmm/src/aspace.rs
@@ -266,7 +266,7 @@ impl AddrSpace {
             }
 
             // This area overlaps with the memory region
-            if area.flags().contains(access_flags) {
+            if !area.flags().contains(access_flags) {
                 return false;
             }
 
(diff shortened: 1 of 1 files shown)


Azure-stars:
The implementation of `axmm` modules may need to serve multiple kernel architectures such as Monolithic kernels and Hypervisors. Previously, we proposed an attempt to set `AddrSpace` in the form of **an independent crate**, but it was not merged. However, some developers agree with this improvement.

You can refer to this design: https://github.com/arceos-hypervisor/axvisor/issues/123. Another reference is https://github.com/arceos-hypervisor/axaddrspace.

modules/axmm/src/aspace.rs
```diff
@@ -248,6 +248,37 @@ impl AddrSpace {
         self.areas.clear(&mut self.pt).unwrap();
     }
 
+    /// Checks whether an access to the specified memory region is valid.
+    ///
+    /// Returns `true` if the memory region given by `range` is all mapped and
+    /// has proper permission flags (i.e. containing `access_flags`).
+    pub fn check_region_access(
```
equation314: Maybe `can_access_range` is a better name?

And change the `range` argument to `start` and `size` for consistency with other functions.

elebirds: okay

equation314 (approved):

PR #254, opened 2025-06-15
feat(mm): add populate_area, fix protect impl

aluojibuzhu:
# Background
# Description
This PR has added the `populate_area` function to update the address detection logic in the `protect` function.  Also, replace manual validation with validate_region.
# Additional notes

### DIFF
--- a/modules/axmm/src/aspace.rs
+++ b/modules/axmm/src/aspace.rs
@@ -2,7 +2,7 @@ use core::fmt;
 
 use axerrno::{AxError, AxResult, ax_err};
 use axhal::mem::phys_to_virt;
-use axhal::paging::{MappingFlags, PageTable, PagingError};
+use axhal::paging::{MappingFlags, PageTable};
 use memory_addr::{
     MemoryAddr, PAGE_SIZE_4K, PageIter4K, PhysAddr, VirtAddr, VirtAddrRange, is_aligned_4k,
 };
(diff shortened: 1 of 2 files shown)


modules/axmm/src/aspace.rs
```diff
@@ -74,6 +74,16 @@ impl AddrSpace {
         Ok(())
     }
 
+    fn validate_region(&self, start: VirtAddr, size: usize) -> AxResult {
```
equation314: `validate_range`

modules/axmm/src/aspace.rs
```diff
@@ -147,17 +150,51 @@ impl AddrSpace {
         Ok(())
     }
 
+    /// Populates the area with physical frames, returning false if the area
+    /// contains unmapped area.
+    pub fn populate_area(&mut self, mut start: VirtAddr, size: usize) -> AxResult {
```
equation314: Why not implement this function in `Backend`?

aluojibuzhu: The populate_area function actually accesses three members of AddrSpace: va_range, areas, and pt. If we implement this function inside the Backend, we would need to pass all these parameters. Wouldn't that be a bit redundant?

equation314: What I mean is that the "poplate" operation is not necessary for `mprotect`. The goal of implementing `protect` in `Bankend` is to handle it differently depending on the kind of `Bankend`.

You can refer to https://github.com/arceos-org/arceos/pull/167.

equation314:
The [spec](https://man7.org/linux/man-pages/man2/mprotect.2.html) of `mprotect` does not say that it will populate the memory region into physical memory.

PR #255, opened 2025-06-15
Always use `-softfloat` target for aarch64 and loongarch64

equation314:
For aarch64 and loongarch64, using a target without `-softfloat` will allow the kernel to use SIMD instructions for some optimization (e.g. `memcpy`). However, properly supporting floating point or SIMD instructions in the kernel would be very cumbersome.

In this PR, we use `-softfloat` target to completely disable kernel FP/SIMD support for aarch64 and loongarch64, regardless of whether the `fp_simd` feature is enabled or not. This means that the `fp_simd` feature will only allow FP/SIMD instructions to be used **in the application, not the kernel**. When the `fp_simd` feature is enabled, only a small piece of extended state switching code is included in the kernel to allow the application using FP/SIMD to run correctly.

However, due to the fact that apps and kernels are linked with the same target in Unikernel, it may cause the app to be unable to use hardware FP/SIMD instructions. But apps in the user mode (monolithic kernel) are not affected. The detailed support table is as the follow:

| FP/SIMD support  | x86_64 | aarch64 | riscv64 | loongarch64 |
|----------|--------|---------|---------|-------------|
| Rust Apps (unikernel) | × | × | × | × |
| C Apps (unikernel) | ✓ | ✓ | × | × |
| Any Apps (monolithic kernel) | ✓ | ✓ | ✓ | ✓ |

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,19 +88,11 @@ include scripts/make/platform.mk
 ifeq ($(ARCH), x86_64)
   TARGET := x86_64-unknown-none
 else ifeq ($(ARCH), aarch64)
-  ifeq ($(findstring fp_simd,$(FEATURES)),)
-    TARGET := aarch64-unknown-none-softfloat
-  else
-    TARGET := aarch64-unknown-none
-  endif
+  TARGET := aarch64-unknown-none-softfloat
 else ifeq ($(ARCH), riscv64)
   TARGET := riscv64gc-unknown-none-elf
 else ifeq ($(ARCH), loongarch64)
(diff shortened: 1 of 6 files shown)


Azure-stars:
LGTM. 

And should we add this modification in https://github.com/arceos-org/axcpu/pull/5?

Azure-stars (approved):

equation314:
> LGTM.
> 
> And should we add this modification in [arceos-org/axcpu#5](https://github.com/arceos-org/axcpu/pull/5)?

Nope

PR #257, opened 2025-06-16
Replace axhal::arch with crate axcpu

equation314:
This PR uses the crate [axcpu](https://github.com/arceos-org/axcpu) to replace architecture-dependent code in `axhal::arch`. It also renames some feature names for consistency (`_` to `-`).

There are some differences from the previous implementation:

#### 1. `SP_EL0` is reset to 0 before each exception handling on aarch64.

This is used to avoid adding extra code in `axcpu` for getting the current task pointer, since we use `SP_EL0` to cache the current task pointer in kernel. 

`SP_EL0` will be corrupted upon entering the user space. To recover it, we first clear `SP_EL0` before entering the exception handler. Once someone reads it for the first time, it will cause a a cache miss and call the slow path to read the actual current task pointer. This logic can be implemented entirely in `axhal`.

#### 2. The current kernel page table root is read and recorded when creating a `TaskContext`, if the `uspace` feature is enabled.

This is used to support the initialization of page table root for kernel threads, but only for kernel threads created at initialization time. After the user program starts running, the user page table is used for both user mode running and kernel mode syscall handling.

This means that if the user program creates a kernel thread in a syscall, it will still use the user page table to initialize `TaskContext`, potentially resulting in undefined behavior. We leave this issue for the future.


modules/axhal/src/platform/loongarch64_qemu_virt/boot.rs
```diff
@@ -65,7 +59,7 @@ unsafe extern "C" fn _start() -> ! {
 
         # Init MMU
```
Azure-stars, line 71: The reason why the `enable_fp` function is not called here is because we are currently using a `softfloat` target, so no SIMD instructions will be generated?

modules/axhal/src/platform/aarch64_common/boot.rs
```diff
@@ -18,88 +16,12 @@ const FLAG_LE: usize = 0b0;
...
-    SCTLR_EL1.modify(SCTLR_EL1::M::Enable + SCTLR_EL1::C::Cacheable + SCTLR_EL1::I::Cacheable);
-    barrier::isb(barrier::SY);
-}
-
 unsafe fn enable_fp() {
-    if cfg!(feature = "fp_simd") {
-        CPACR_EL1.write(CPACR_EL1::FPEN::TrapNothing);
-        barrier::isb(barrier::SY);
-    }
+    // FP/SIMD needs to be enabled early, as the compiler may generate SIMD
```
Azure-stars, line 20: And is this comment necessary when we always use `-softfloat` target?

modules/axhal/src/platform/loongarch64_qemu_virt/boot.rs
equation314: Yes

equation314: I think it's better to enable FP and LSX before `clear_bss`, to make the platform crate be easier reused by other projects (single-thread, no-IRQ kernel can safely use targets without `softfloat`).

I'll update it later.

Azure-stars: LGTM.

Azure-stars (dismissed):

Azure-stars:
Should we need to rebase the commits to remove the CI-fail record?

equation314:
Nope, they won't show up after the final merge.

Azure-stars (approved):

PR #267, opened 2025-06-21
[fix]: incorrect writev implementation

numpy1314:
The original code assumes that each sys_write must be completely successful and does not handle:
1、The system call returns an error
2、Partial write (actual number of bytes written < number of bytes requested)

The new implementation complies with the standard:
The modified logic strictly follows the specification of the UNIX writev(2) system call:
If an error occurs, it will be terminated immediately and an error code will be returned
When a partial write occurs, the amount of data that has been successfully written will be retained and returned
The final return value ret accurately represents the total number of bytes actually written​

### DIFF
--- a/api/arceos_posix_api/src/imp/io.rs
+++ b/api/arceos_posix_api/src/imp/io.rs
@@ -66,7 +66,15 @@ pub unsafe fn sys_writev(fd: c_int, iov: *const ctypes::iovec, iocnt: c_int) -> 
         let iovs = unsafe { core::slice::from_raw_parts(iov, iocnt as usize) };
         let mut ret = 0;
         for iov in iovs.iter() {
-            ret += write_impl(fd, iov.iov_base, iov.iov_len)?;
+            let result = sys_write(fd, iov.iov_base, iov.iov_len);
+            if result < 0 {
+                return Ok(result);
+            }
+            ret += result;
+
+            if result < iov.iov_len as isize {
+                break;
+            }
         }
 
         Ok(ret)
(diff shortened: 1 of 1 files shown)


api/arceos_posix_api/src/imp/io.rs
```diff
@@ -66,7 +66,15 @@ pub unsafe fn sys_writev(fd: c_int, iov: *const ctypes::iovec, iocnt: c_int) ->
         let iovs = unsafe { core::slice::from_raw_parts(iov, iocnt as usize) };
         let mut ret = 0;
         for iov in iovs.iter() {
-            ret += write_impl(fd, iov.iov_base, iov.iov_len)?;
+            let result = sys_write(fd, iov.iov_base, iov.iov_len);
```
equation314: Why change back to `sys_write`?

AsakuraMizu:
This is still incorrect, though. You may refer to https://github.com/oscomp/arceos/pull/28#issuecomment-2780653153 and https://github.com/oscomp/starry-next/pull/47/files#diff-afec6b20b72a97221006fe9f906ba75b011436511cca0b53674ebb1542319c17R48.

numpy1314:
> 但这仍然是错误的。您可以参考[oscomp#28 (评论)](https://github.com/oscomp/arceos/pull/28#issuecomment-2780653153)和https://github.com/oscomp/starry-next/pull/47/files#diff-afec6b20b72a97221006fe9f906ba75b011436511cca0b53674ebb1542319c17R48。

If I choose to use continue to skip the loop when iov_len is 0, does this conform to the logic of Linux?

api/arceos_posix_api/src/imp/io.rs
numpy1314: Since I haven't thought of a suitable way to use write_impl yet, I will change it back to using write_impl later.

numpy1314: This part has been modified in the latest commit. I forgot to check the previous commits because they were done separately.

equation314 (approved):

PR #269, opened 2025-06-24
[fix] AddrSpace clone && fdon't map to zero address for lazily allocated memory area 

aluojibuzhu:
# Background
Partial modifications are based on the[[feat] Support for ext4 junior ](https://github.com/arceos-org/arceos/commit/111dbc27acbc7e445767d288c2c08efb25f1e401)support.
# Description
- fix ：AddrSpace::clone_or_err cannot handle allocation mappings with populate=false by skipping unmapped pages in the source area and trying to allocate pages for unmapped pages in the destination area.
And some minor adjustments.

### DIFF
--- a/modules/axmm/src/aspace.rs
+++ b/modules/axmm/src/aspace.rs
@@ -315,6 +315,11 @@ impl AddrSpace {
                 .areas
                 .map(new_area, &mut new_aspace.pt, false)
                 .map_err(mapping_err_to_ax_err)?;
+
+            if matches!(backend, Backend::Linear { .. }) {
+                continue;
+            }
+
             // Copy data from old memory area to new memory area.
             for vaddr in
                 PageIter4K::new(area.start(), area.end()).expect("Failed to create page iterator")
--- a/modules/axmm/src/aspace.rs
+++ b/modules/axmm/src/aspace.rs
@@ -302,22 +302,20 @@ impl AddrSpace {
         false
     }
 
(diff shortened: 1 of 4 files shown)


modules/axmm/src/aspace.rs
```diff
@@ -268,6 +301,60 @@ impl AddrSpace {
         }
         false
     }
+
+    /// Clone a [`AddrSpace`] by re-mapping all [`MemoryArea`]s in a new page table and copying data in user space.
+    pub fn clone_or_err(&mut self) -> AxResult<Self> {
```
equation314, line 306: Rename to `dup` to distinguish from `Clone::clone`

modules/axmm/src/aspace.rs
```diff
@@ -268,6 +301,60 @@ impl AddrSpace {
...
+                .areas
+                .map(new_area, &mut new_aspace.pt, false)
+                .map_err(mapping_err_to_ax_err)?;
+
+            if matches!(backend, Backend::Linear { .. }) {
+                continue;
+            }
+
+            // Copy data from old memory area to new memory area.
+            for vaddr in
```
equation314, line 324: Distribute implementations to the backend, e.g., move the following lines into `Backend::dup_alloc`.

modules/axmm/src/aspace.rs
AsakuraMizu: FYI, it was [later renamed to](https://github.com/oscomp/arceos/blob/main/modules/axmm/src/aspace.rs#L563C12-L563C21) `try_clone`, which better reflects the semantics that it may fail.

PR #270, opened 2025-06-26
[feat] Enhance dyn driver

ZR233:
feat(axdriver): 

- Enhanced dynamic driver probing with dtb.
- Driver registration method similar to Linux module_driver.
- Add gic and block dyn driver.
- Supports dependency injection, obtaining drivers of the specified type through get::<T:Class> and get_list::<T:Class>.
- Supports type conversion through `downcast`, which allows converting a `Class` driver to a specific type. For example, `Device<Intc>.downcast -> Device<Gicv3>`.

test cmd:
```shell
make A=examples/shell ARCH=aarch64 LOG=debug FEATURES=driver-dyn LOG=info BUS=mmio BLK=y run
```

### DIFF
--- a/modules/axdriver/Cargo.toml
+++ b/modules/axdriver/Cargo.toml
@@ -49,6 +49,7 @@ axdma = { workspace = true, optional = true }
 rdrive = { version = "0.15", optional = true }
 axerrno = "0.1"
 lazyinit = "0.2"
+memory_addr = "0.4"
 
 [target.'cfg(target_arch = "aarch64")'.dependencies]
 arm-gic-driver = { version = "0.14", optional = true }
--- a/modules/axdriver/Cargo.toml
+++ b/modules/axdriver/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/arceos-org/arceos/tree/main/modules/axdriver"
 documentation = "https://arceos-org.github.io/arceos/axdriver/index.html"
 
 [features]
-dyn = []
+dyn = ["dep:rdrive", "dep:arm-gic-driver"]
 bus-mmio = []
 bus-pci = ["dep:axdriver_pci", "dep:axhal", "dep:axconfig"]
 net = ["axdriver_net"]
@@ -46,3 +46,9 @@ axalloc = { workspace = true, optional = true }
 axhal = { workspace = true, optional = true }
 axconfig = { workspace = true, optional = true }
 axdma = { workspace = true, optional = true }
+rdrive = { version = "0.15", optional = true }
+axerrno = "0.1"
+lazyinit = "0.2"
+
+[target.'cfg(target_arch = "aarch64")'.dependencies]
+arm-gic-driver = { version = "0.14", optional = true }
--- a/modules/axdriver/src/dyn_drivers/blk/mod.rs
+++ b/modules/axdriver/src/dyn_drivers/blk/mod.rs
@@ -1,0 +1,87 @@
+use axdriver_base::{BaseDriverOps, DevError, DevResult, DeviceType};
+use axdriver_block::BlockDriverOps;
+use rdrive::{Device, driver::block::io};
+
+#[cfg(feature = "virtio-blk")]
+mod virtio;
+
+pub struct Block(Device<rdrive::driver::Block>);
+
+impl BaseDriverOps for Block {
+    fn device_type(&self) -> DeviceType {
+        DeviceType::Block
+    }
+    fn device_name(&self) -> &str {
+        self.0.descriptor().name
+    }
+}
+
+impl BlockDriverOps for Block {
+    fn num_blocks(&self) -> u64 {
+        self.0.lock().unwrap().num_blocks() as _
+    }
+    fn block_size(&self) -> usize {
+        self.0.lock().unwrap().block_size()
+    }
+    fn flush(&mut self) -> DevResult {
+        self.0
+            .lock()
+            .unwrap()
+            .flush()
+            .map_err(maping_io_err_to_dev_err)
+    }
+
+    fn read_block(&mut self, block_id: u64, buf: &mut [u8]) -> DevResult {
+        self.0
+            .lock()
+            .unwrap()
+            .read_block(block_id as _, buf)
+            .map_err(maping_io_err_to_dev_err)
+    }
+
+    fn write_block(&mut self, block_id: u64, buf: &[u8]) -> DevResult {
+        self.0
+            .lock()
+            .unwrap()
+            .write_block(block_id as _, buf)
+            .map_err(maping_io_err_to_dev_err)
+    }
+}
+
+impl From<Device<rdrive::driver::Block>> for Block {
+    fn from(base: Device<rdrive::driver::Block>) -> Self {
+        Self(base)
+    }
+}
+
+fn maping_io_err_to_dev_err(err: io::Error) -> DevError {
+    match err.kind {
+        io::ErrorKind::Other(_error) => DevError::Io,
+        io::ErrorKind::NotAvailable => DevError::BadState,
+        io::ErrorKind::BrokenPipe => DevError::BadState,
+        io::ErrorKind::InvalidParameter { name: _ } => DevError::InvalidParam,
+        io::ErrorKind::InvalidData => DevError::InvalidParam,
+        io::ErrorKind::TimedOut => DevError::Io,
+        io::ErrorKind::Interrupted => DevError::Again,
+        io::ErrorKind::Unsupported => DevError::Unsupported,
+        io::ErrorKind::OutOfMemory => DevError::NoMemory,
+        io::ErrorKind::WriteZero => DevError::InvalidParam,
+    }
+}
+
+fn maping_dev_err_to_io_err(err: DevError) -> io::Error {
+    let kind = match err {
+        DevError::Again => io::ErrorKind::Interrupted,
+        DevError::AlreadyExists => io::ErrorKind::Other("Already exists".into()),
+        DevError::BadState => io::ErrorKind::BrokenPipe,
+        DevError::InvalidParam => io::ErrorKind::InvalidData,
+        DevError::Io => io::ErrorKind::Other("I/O error".into()),
+        DevError::NoMemory => io::ErrorKind::OutOfMemory,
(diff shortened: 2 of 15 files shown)


modules/axdriver/src/dyn_drivers/mod.rs
```diff
@@ -0,0 +1,37 @@
+use core::{error::Error, ptr::NonNull};
+
+use alloc::{boxed::Box, format};
+use axhal::{
+    mem::{MemoryAddr, PhysAddr, phys_to_virt},
+    paging::MappingFlags,
+};
+use axmm::AxError;
```
Azure-stars: Use [axerrno](https://crates.io/crates/axerrno) instead of `axmm`

modules/axdriver/src/lib.rs
```diff
@@ -147,6 +159,14 @@ impl AllDevices {
     }
 }
 
+/// Sets up the device driver subsystem.
+pub fn setup(dtb: usize) {
+    let dtb_virt = phys_to_virt(dtb.into());
+    if let Some(dtb) = core::ptr::NonNull::new(dtb_virt.as_mut_ptr()) {
```
Azure-stars: If it return `None`, then `dtb_virt` is 0. But when will `dtb_virt` be 0? Maybe we can directly return when `dtb` is 0?

modules/axmm/src/lib.rs
```diff
@@ -12,7 +12,7 @@ mod backend;
 pub use self::aspace::AddrSpace;
 pub use self::backend::Backend;
 
-use axerrno::{AxError, AxResult};
+pub use axerrno::{AxError, AxResult};
```
Azure-stars: Keep it as private

modules/axruntime/Cargo.toml
```diff
@@ -17,6 +17,7 @@ irq = ["axhal/irq", "axtask?/irq", "percpu", "kernel_guard"]
 tls = ["axhal/tls", "axtask?/tls"]
 alloc = ["axalloc"]
 paging = ["axhal/paging", "axmm"]
+driver = ["axdriver"]
```
Azure-stars: Maybe it can rename as `driver-dyn` to represent dynamic drivers that should be set up at runtime?

modules/axdriver/src/bus/mod.rs
Azure-stars, line 1: Maybe when `dyn` feature is enabled, this mod will not be used? So we can ignore it in `lib.rs` of the root and remove `#[allow(unused]` here.

modules/axdriver/Cargo.toml
```diff
@@ -46,3 +46,8 @@ axalloc = { workspace = true, optional = true }
 axhal = { workspace = true, optional = true }
 axconfig = { workspace = true, optional = true }
 axdma = { workspace = true, optional = true }
+rdrive = { version = "0.15", optional = true }
+axmm = { workspace = true, optional = true }
```
Azure-stars: remove it and use `axerrno`

modules/axdriver/src/dyn_drivers/blk/virtio.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    let dev = Device::try_new(transport).map_err(|e| {
+        OnProbeError::other(format!(
+            "failed to initialize Virtio Block device at [PA:{:?},): {:?}",
+            mmio_base, e
+        ))
+    })?;
+
+    let dev = BlockDivce(dev);
+
```
Azure-stars: remove some line break?

modules/axdriver/src/dyn_drivers/blk/virtio.rs
```diff
@@ -0,0 +1,112 @@
...
+    ],
+);
+
+fn probe(info: FdtInfo<'_>, plat_dev: PlatformDevice) -> Result<(), OnProbeError> {
+    let mut reg = info.node.reg().ok_or(OnProbeError::other(alloc::format!(
+        "[{}] has no reg",
+        info.node.name()
+    )))?;
+
+    let base_reg = reg.next().unwrap();
```
Azure-stars: Can we return an error instead of `unwrap`?

modules/axdriver/src/dyn_drivers/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+#[allow(unused)]
+/// maps a mmio physical address to a virtual address.
+fn iomap(addr: PhysAddr, size: usize) -> Result<NonNull<u8>, Box<dyn Error>> {
+    let end = (addr.as_usize() + size).align_up_4k();
+    let start = addr.align_down_4k();
+    let size = end - start.as_usize();
+
+    let start_virt = phys_to_virt(start);
+
+    match axmm::kernel_aspace().lock().map_linear(
```
Azure-stars: I don't suggest import `mm` module to `driver` module directly. This will cause the driver module to introduce unnecessary mapping functions. Perhaps we can take one of three approaches:
1. Return a `iomap_areas` like `axhal::mem::memory_regions()` and map them in `axruntime`.
2. Introduce a trait, define the functions required for address mapping, call the trait method in `iomap`, and instantiate the trait at `runtime`.
3. Use `crate_interface`(Don't recommend)

ZR233: I prefer option 2 or 3. The drivers inside the probe function may depend on functionalities provided by other drivers. If we use approach 1, earlier drivers might not be available yet because they haven’t been mapped, making them unusable.

modules/axdriver/src/dyn_drivers/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+use axhal::mem::{MemoryAddr, PhysAddr, phys_to_virt};
+
+mod intc;
+
+#[cfg(feature = "block")]
+pub mod blk;
+
+#[allow(unused)]
+/// maps a mmio physical address to a virtual address.
+fn iomap(addr: PhysAddr, size: usize) -> Result<NonNull<u8>, Box<dyn Error>> {
```
Azure-stars, line 38: Why return `Box<dyn Error>` as error type?

Maybe you just return a string? Can we return `AxError`?

modules/axdriver/src/lib.rs
```diff
@@ -78,11 +79,18 @@ mod virtio;
...
+mod dyn_drivers;
+
 pub mod prelude;
 
 #[allow(unused_imports)]
 use self::prelude::*;
 pub use self::structs::{AxDeviceContainer, AxDeviceEnum};
+use axerrno::AxResult;
+use axhal::mem::phys_to_virt;
+use axhal::mem::{PhysAddr, VirtAddr};
```
Azure-stars: merge Line 91 and 92

modules/axdriver/src/structs/static.rs
```diff
@@ -73,3 +73,5 @@ impl<D> Default for AxDeviceContainer<D> {
         Self(Default::default())
     }
 }
+
+pub fn setup(_dtb: core::ptr::NonNull<u8>) {}
```
Azure-stars: Is there any possible work in the future that requires initializing the static type driver at runtime?

If not, is it possible to define `setup` only in the `dyn` type and **export this method** in the `lib.rs` instead of defining it directly in `lib.rs`?

modules/axdriver/src/lib.rs
```diff
@@ -91,6 +99,11 @@ pub use self::structs::AxDisplayDevice;
 #[cfg(feature = "net")]
 pub use self::structs::AxNetDevice;
 
+/// A function type that maps a physical address to a virtual address. map flags should be read/write/device.
+pub type IoMapFunc = fn(PhysAddr, usize) -> AxResult<VirtAddr>;
+
+static IO_MAP_FUNC: LazyInit<IoMapFunc> = LazyInit::new();
```
Azure-stars: move it to `dyn_drivers` mod

modules/axdriver/src/dyn_drivers/blk/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+        io::ErrorKind::TimedOut => DevError::Io,
+        io::ErrorKind::Interrupted => DevError::Again,
+        io::ErrorKind::Unsupported => DevError::Unsupported,
+        io::ErrorKind::OutOfMemory => DevError::NoMemory,
+        io::ErrorKind::WriteZero => DevError::InvalidParam,
+    }
+}
+
+fn deverror_to_io_error(err: DevError) -> io::Error {
+    match err {
```
Azure-stars:
```rust
        let kind = match err {
            DevError::Again => Interrupted,
            DevError::AlreadyExists => Other("Already exists".into()),
            DevError::BadState => BrokenPipe,
            DevError::InvalidParam => InvalidData,
            DevError::Io => Other("I/O error".into()),
            DevError::NoMemory => OutOfMemory,
            DevError::ResourceBusy => Other("Resource busy".into()),
            DevError::Unsupported => Unsupported,
        };
        io::Error { kind, success_pos: 0 }
```

modules/axdriver/src/dyn_drivers/blk/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+impl From<Device<rdrive::driver::Block>> for Block {
+    fn from(base: Device<rdrive::driver::Block>) -> Self {
+        Self(base)
+    }
+}
+
+fn io_error_to_deverror(err: io::Error) -> DevError {
```
Azure-stars: `io_err_to_dev_err`. see: https://github.com/arceos-org/arceos/blob/65e97ac15ca6be406745363416668e1b7df6d037/modules/axmm/src/lib.rs#L24-L32

modules/axdriver/src/dyn_drivers/mod.rs
ZR233: `OnProbeError` uses `Box<dyn Error>`, these two error type are in other crates, if we want to use `?` to convert err, need to create new type to warp a String or AxError，and boxed error has more err msg than `AxError`, make it easier to find error location.

equation314:
This PR is scheduled to be included in the next release window (0.3.0, expected 9/30), and PRs marked 0.2.0 will be processed first.

PR #277, opened 2025-07-03
[CI] add check for build and doc CI tasks under macOS

Azure-stars:
This PR adds check for building and checking documentation of ArceOS on the `macOS` environment in CICD.

And it fixes two bugs that caused `ArceOS` to fail to build on `macOS`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,11 +38,10 @@ jobs:
       run: make clippy ARCH=${{ matrix.arch }}
 
   build:
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
         arch: [x86_64, riscv64, aarch64, loongarch64]
         rust-toolchain: [nightly, nightly-2025-05-20]
     env:
@@ -86,11 +85,10 @@ jobs:
       run: make ARCH=${{ matrix.arch }} A=examples/httpserver-c
 
   build-for-other-platforms:
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
         rust-toolchain: [nightly, nightly-2025-05-20]
     env:
       RUSTUP_TOOLCHAIN: ${{ matrix.rust-toolchain }}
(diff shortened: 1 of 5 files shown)


Azure-stars:
The implement in [axns::link](https://github.com/arceos-org/arceos/blob/feat/macos_ci/modules/axns/src/link.rs) refer to [linkme](https://github.com/dtolnay/linkme/blob/master/impl/src/linker.rs).

modules/axns/src/lib.rs
```diff
@@ -23,13 +23,11 @@ extern crate alloc;
...
-
 use lazyinit::LazyInit;
 
-unsafe extern "C" {
-    fn __start_axns_resource();
-    fn __stop_axns_resource();
-}
+mod link;
+
+pub use link::*;
```
equation314: The following codes all use `link::`, can this line be removed?

modules/axns/src/lib.rs
```diff
@@ -233,14 +231,9 @@ macro_rules! def_resource {
 
             impl $name {
                 unsafe fn deref_from_base(&self, ns_base: *mut u8) -> &$ty {
-                    unsafe extern {
-                        fn __start_axns_resource();
-                    }
-
-                    #[unsafe(link_section = "axns_resource")]
-                    static RES: $ty = $default;
+                    $crate::def_static_resource!($ty, $default);
```
equation314: Use `def_static_resource!(RES, $ty, $default)`, and remove `res_ptr()` in the macro.

equation314:
How about reducing the number of tests under macOS? Because the concurrent job limit (5 for free) is reached very quickly and the CI results are slow to come out: https://docs.github.com/en/actions/concepts/overview/usage-limits-billing-and-administration.

Azure-stars:
Now I put necessary checks in a single job for macOS. Here we will check CI results based on MacOS's default target. 

And we have reduced the number of apps being built, focusing on one or two cases. By varying feature flags and PLATFORM configurations, we aim to verify build correctness while keeping the CI job lightweight.

Most potential errors related to the bare-metal toolchain like `x86_64-unknown-none` may have already been caught by previous jobs.

equation314 (dismissed):

.github/workflows/build.yml
```diff
@@ -151,3 +149,53 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        rust-toolchain: [nightly, nightly-2025-05-20]
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+      with:
+        toolchain: ${{ matrix.rust-toolchain }}
+        components: rust-src, clippy
+        targets: x86_64-unknown-none, riscv64gc-unknown-none-elf, aarch64-unknown-none-softfloat, loongarch64-unknown-none-softfloat
```
equation314: Remove unused targets as we only build on `x86_64-unknown-none`

equation314 (approved):

PR #281, opened 2025-07-21
[feat] support arceos virtualization

Josen-B:


aarkegz:
This PR is good, I think. Could you please cherry-pick the original commits to preserve the original commit info? @Josen-B

modules/axtask/src/task.rs
```diff
@@ -398,6 +404,22 @@ impl TaskInner {
...
+    #[cfg(feature = "smp")]
+    #[inline]
+    pub fn cpu_id(&self) -> u32 {
+        self.cpu_id.load(Ordering::Acquire)
+    }
+
+    /// Set the CPU ID where the task is running or will run.
+    #[cfg(feature = "smp")]
+    #[inline]
+    pub fn set_cpu_id(&self, cpu_id: u32) {
```
equation314, line 419: Others should not modify the CPU ID of a task, make this function `pub(crate) `.

Josen-B:
> This PR is good, I think. Could you please cherry-pick the original commits to preserve the original commit info? @Josen-B

Ok, i cherry-pick the original commits and repull requests in https://github.com/arceos-org/arceos/pull/282

modules/axtask/src/task.rs
Josen-B: ok，modify this in https://github.com/arceos-org/arceos/pull/282

PR #287, opened 2025-08-14
[feat] make irq_handler public for hypervisor reuse

ZR233:


modules/axhal/src/irq.rs
```diff
@@ -4,8 +4,13 @@ use axcpu::trap::{IRQ, register_trap_handler};
 
 pub use axplat::irq::{handle, register, set_enable, unregister};
 
+/// IRQ handler.
+///
+/// # Warn
+///
+/// Make sure called in an interrupt context or hypervisor context.
```
equation314: `hypervisor context` -> `hypervisor VM exit handler`

ZR233: Modified

equation314 (approved):

PR #296, opened 2025-11-05
tlb: kernel after write tlb need flush

zgyxyd666:
Without a flush, stale TLB entries may still reference previous mappings, causing
incorrect address translations during early kernel initialization.

## Change Details
Add `axhal::asm::flush_tlb(None);` immediately after `write_kernel_page_table()`.


modules/axmm/src/lib.rs
```diff
@@ -106,6 +106,8 @@ pub fn init_memory_management() {
     debug!("kernel address space init OK: {:#x?}", kernel_aspace);
     KERNEL_ASPACE.init_once(SpinNoIrq::new(kernel_aspace));
     unsafe { axhal::asm::write_kernel_page_table(kernel_page_table_root()) };
+    // flush all tlb
+    axhal::asm::flush_tlb(None);
```
equation314, line 89: Also add it in `init_memory_management_secondary`. And the flush is not required on x86_64.

equation314 (commented):
Yes, TLB flushing was missed after https://github.com/arceos-org/arceos/pull/257. Thanks!

PR #299, opened 2025-11-18
Fix: `run_cmd` unsoundness and clippy warnings in the example `shell`

### LINKED ISSUE #204: Unsoundness in run_cmd
## Description

The `run_cmd` function https://github.com/arceos-org/arceos/blob/82d9a05b3b404d13e99893d683f41519b667d4e8/examples/shell/src/cmd.rs#L275 appears to have a potential unsoundness issue when processing invalid UTF-8 sequences. Specifically, the function uses `unsafe { core::str::from_utf8_unchecked(line) }` to convert a byte slice (`line`) into a string without verifying that the input is valid UTF-8. If the input contains invalid UTF-8 sequences, this can lead to undefined behavior (UB), as shown in the provided PoC.
```
pub fn run_cmd(line: &[u8]) {
    let line_str = unsafe { core::str::from_utf8_unchecked(line) };
    let (cmd, args) = split_whitespace(line_str);
    if !cmd.is_empty() {
        for (name, func) in CMD_TABLE {
            if cmd == *name {
                func(args);
                return;
            }
        }
        println!("{}: command not found", cmd);
    }
}
```
## Steps to Reproduce

Here's a minimal reproducible example that demonstrates the issue:
```
fn main() {
    let invalid_utf8: &[u8] = &[0x80, 0x81, 0x82];
    run_cmd(invalid_utf8);
}
```
output：
```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.36s
     Running `target\debug\lwz.exe`
thread 'main' panicked at core\src\panicking.rs:221:5:
unsafe precondition(s) violated: hint::unreachable_unchecked must never be reached
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread caused non-unwinding panic. aborting.
error: process didn't exit successfully: `target\debug\lwz.exe` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)

```
## Suggested Fix
1. The unsoundness can be mitigated by replacing the use of core::str::from_utf8_unchecked with the safer core::str::from_utf8. This change ensures that the function validates the input before attempting the conversion.
2. Mark run_cmd as 'unsafe' function and write the safety precondition.

zjp-CN:
This PR focuses on the shell example:
* closes https://github.com/arceos-org/arceos/issues/204
* fixes clippy warnings
* handles `.unwrap()` calls

### DIFF
--- a/examples/shell/src/cmd.rs
+++ b/examples/shell/src/cmd.rs
@@ -87,16 +87,7 @@ fn do_ls(args: &str) {
         let file_type = metadata.file_type();
         let file_type_char = file_type_to_char(file_type);
         let rwx = file_perm_to_rwx(metadata.permissions().mode());
-        let Ok(rwx) = str::from_utf8(&rwx) else {
-            #[cfg(feature = "axstd")]
-            let err = io::Error::InvalidData;
-            #[cfg(not(feature = "axstd"))]
-            let err = io::Error::new(
-                io::ErrorKind::InvalidData,
-                "The rwx value is not utf8 encoded.",
-            );
-            return Err(err);
-        };
+        let rwx = str::from_utf8(&rwx).unwrap();
         println!("{file_type_char}{rwx} {size:>8} {entry}");
         Ok(())
     }
--- a/examples/shell/src/cmd.rs
+++ b/examples/shell/src/cmd.rs
@@ -84,7 +84,7 @@ fn do_ls(args: &str) {
         let file_type_char = file_type_to_char(file_type);
         let rwx = file_perm_to_rwx(metadata.permissions().mode());
         let rwx = core::str::from_utf8(&rwx).unwrap();
-        println!("{}{} {:>8} {}", file_type_char, rwx, size, entry);
+        println!("{file_type_char}{rwx} {size:>8} {entry}");
         Ok(())
     }
 
@@ -95,7 +95,7 @@ fn do_ls(args: &str) {
         }
 
         if print_name {
-            println!("{}:", name);
+            println!("{name}:");
         }
         let mut entries = fs::read_dir(name)?
             .filter_map(|e| e.ok())
@@ -176,7 +176,7 @@ fn do_echo(args: &str) {
             print_err!("echo", fname, e);
         }
     } else {
-        println!("{}", args)
+        println!("{args}")
     }
 }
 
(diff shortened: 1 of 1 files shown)


examples/shell/src/cmd.rs
```diff
@@ -279,7 +273,7 @@ fn do_exit(_args: &str) {
 }
 
 pub fn run_cmd(line: &[u8]) {
-    let line_str = unsafe { core::str::from_utf8_unchecked(line) };
+    let line_str = core::str::from_utf8(line).unwrap();
```
equation314: Print an error message instead of calling `unwrap` if `from_utf8` returns `Err`.

zjp-CN: I'll do that.

There are some other `.unwrap()` in the same file. Should they be fixed too?

https://github.com/arceos-org/arceos/blob/c5a2cecbdbaf3253bdbf29bb80c2139e6555e7af/examples/shell/src/cmd.rs#L72

equation314: This place seems to never fail.

zjp-CN: From the API doc:
> Possible cases for error:
> * Current directory does not exist.
> * There are insufficient permissions to access the current directory.

https://doc.rust-lang.org/std/env/fn.current_dir.html#errors

zjp-CN: > This place seems to never fail.

I see what you mean here. Yeah, it's true in ArceOS.

https://github.com/arceos-org/arceos/blob/c5a2cecbdbaf3253bdbf29bb80c2139e6555e7af/modules/axfs/src/root.rs#L279-L281

Anyway, since it's also works on the Rust standard library, I'll handle the error.

examples/shell/src/cmd.rs
```diff
@@ -83,7 +87,16 @@ fn do_ls(args: &str) {
         let file_type = metadata.file_type();
         let file_type_char = file_type_to_char(file_type);
         let rwx = file_perm_to_rwx(metadata.permissions().mode());
-        let rwx = core::str::from_utf8(&rwx).unwrap();
+        let Ok(rwx) = str::from_utf8(&rwx) else {
```
equation314: Converting from "rwx" to a string won't fail, just use `unwrap`.

equation314 (approved):

PR #301, opened 2025-11-19
Fix: Add direct optional axhal dependency to axdisplay, axfs, and axsync

### LINKED ISSUE #300: Some crates should expose indirect axhal as a direct optional dependency to compile on multiple platforms
For crates that transively/indirectly rely on axhal, there is no way to compile them on non-host targets/platforms, like axfs:

```text
arceos/modules/axfs $ cargo b --target aarch64-unknown-none-softfloat
   Compiling axhal v0.2.0 (/home/gh-zjp-CN/arceos/arceos/modules/axhal)
error[E0463]: can't find crate for `axplat_aarch64_qemu_virt`
  --> modules/axhal/src/lib.rs:48:9
   |
48 |         extern crate axplat_aarch64_qemu_virt;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate

For more information about this error, try `rustc --explain E0463`.
error: could not compile `axhal` (lib) due to 1 previous error
```

The solution is to add optional axhal in axfs's Cargo.toml, and compile with `-Faxhal/defplat`

```diff
[dependencies]
+axhal = { workspace = true, optional = true }
```

```bash
arceos/modules/axfs $ cargo b --target aarch64-unknown-none-softfloat -Faxhal/defplat 
```

zjp-CN:
* fixes https://github.com/arceos-org/arceos/issues/300 

These crates didn't directly rely on axhal (or axfeat, or axstd). 
So the fix allows axhal's defplat feature and make them compile on non-host targets respectively.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -317,6 +317,7 @@ version = "0.2.0"
 dependencies = [
  "axdriver",
  "axdriver_display",
+ "axhal",
  "axsync",
  "lazyinit",
  "log",
@@ -448,6 +449,7 @@ dependencies = [
  "axfs_devfs",
  "axfs_ramfs",
  "axfs_vfs",
+ "axhal",
  "axio",
  "axns",
  "axsync",
@@ -814,6 +816,7 @@ dependencies = [
 name = "axsync"
 version = "0.2.0"
 dependencies = [
+ "axhal",
  "axsync",
  "axtask",
  "kspin",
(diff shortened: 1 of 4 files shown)


zjp-CN:
Close the PR as it's not a desired fix. But I'll leave the issue open.

PR #304, opened 2025-12-03
fix(tests): resolve test_ramfs failure when dyn feature is enabled

ctrlshiftm129:


AsakuraMizu:
It's not a good idea to add an unused feature just because some tests fail.

PR #309, opened 2025-12-10
get bootargs message

Josen-B:


modules/axhal/src/dtb.rs
```diff
@@ -0,0 +1,52 @@
...
+    let fdt_bytes = unsafe {
+        let ptr = virt_addr as *const u8;
+        let size = fdt_header.total_size() as usize;
+        core::slice::from_raw_parts(ptr, size)
+    };
+
+    let fdt = match Fdt::from_bytes(fdt_bytes) {
+        Ok(fdt) => fdt,
+        Err(_) => return None,
+    };
```
AsakuraMizu: Would it be better to cache the entire FDT structure? It would also provide an interface for possible future changes.

modules/axhal/src/dtb.rs
```diff
@@ -0,0 +1,52 @@
+//! DTB (Device Tree Blob) related functionality.
+use fdt_parser::{Fdt, FdtHeader};
+
+use core::fmt::Write;
+use lazyinit::LazyInit;
+
+static BOOTARGS_STR: LazyInit<heapless::String<256>> = LazyInit::new();
```
AsakuraMizu: Cache unused?

AsakuraMizu: Storing a `'static str` might also work.

modules/axhal/src/dtb.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// Get the bootargs from the device tree.
+pub fn bootargs_message() -> Option<&'static str> {
```
AsakuraMizu: [Nitpick] Name suggestion: `get_xxx`?

modules/axhal/Cargo.toml
```diff
@@ -68,6 +68,7 @@ log.workspace = true
 memory_addr.workspace = true
 page_table_multiarch = { workspace = true, optional = true }
 percpu.workspace = true
+fdt-parser = "0.4"
```
AsakuraMizu, line 71: What's the difference between version 0.4 and 0.5? I see that 0.5.x is already published.

modules/axhal/src/dtb.rs
```diff
@@ -0,0 +1,52 @@
+//! DTB (Device Tree Blob) related functionality.
+use fdt_parser::{Fdt, FdtHeader};
+
+use core::fmt::Write;
+use lazyinit::LazyInit;
+
```
AsakuraMizu, line 5: Perhaps it would be better to move `BOOT_ARG` and the related `get_bootarg` to this module and rename them to something like `xxx_DTB_xxx`.

modules/axhal/src/dtb.rs
Josen-B: that's right

modules/axhal/src/dtb.rs
Josen-B: `CACHED_FDT` is used

modules/axhal/src/dtb.rs
Josen-B: rename `bootargs_message`  to `get_chosen`

modules/axhal/Cargo.toml
Josen-B: 0.5 has not bootargs() and chosen(), it use find_node() to get this message. I think 0.4 is more convenient

modules/axhal/src/dtb.rs
Josen-B: I think so, moved.

PR #314, opened 2025-12-17
Introduce `max-cpu-num` config item, remove all direct usages of `axconfig::plat::CPU_NUM`, use `axplat` to get cpu num.

aarkegz:
Currently ArceOS uses `axconfig::plat::CPU_NUM` to determine the number of CPUs in the target system statically, which is straightforward but quite inflexible. The number of CPUs should be detected and reported by platform crates during initialization. And another platform-independent constant, specifying the maximum number of CPUs supported by ArceOS binary, should be used to determine the size of the bootstack section, the percpu section, etc.

In this PR:

1. A new config item `max-cpu-num` is introduced, which is translated into `axconfig::MAX_CPU_NUM`.
2. Methods and static variables are added to `axhal` to read the number of CPUs during initialization, using `axplat::power::cpu_num()` introduced in https://github.com/arceos-org/axplat_crates/pull/29 . They also handle the abnormal scenarios where platform provided `cpu_num` is larger than `axconfig::MAX_CPU_NUM`.
3. All direct usages of `axconfig::plat::CPU_NUM` are removed and replaced with `axhal::cpu_num()` or `axconfig::MAX_CPU_NUM`.
4. The way to determine and build a full `AxCpuMask` is updated.

Possible issues:

1. `cpu-num` should no longer be a required config item. However, dynamic platforms must still include a dummy greater-than-one `cpu-num` in the config to tell the ArceOS build system that the feature `smp` should be enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,3 +108,7 @@ page_table_multiarch = "0.5"
 percpu = "0.2"
 scope-local = "0.1"
 spin = "0.10"
+
+[patch."https://github.com/arceos-org/axplat_crates.git"]
+axplat = { path = "../axplat_crates/axplat" }
+axplat-x86-pc = { path = "../axplat_crates/platforms/axplat-x86-pc" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,13 +108,3 @@ page_table_multiarch = "0.5"
 percpu = "0.2"
 scope-local = "0.1"
 spin = "0.10"
-
-# [patch."https://github.com/arceos-org/axplat_crates.git"]
-# axplat = { git = "https://github.com/arceos-org/axplat_crates.git", branch = "feat/cpu_num" }
-# axplat-aarch64-bsta1000b = { git = "https://github.com/arceos-org/axplat_crates.git", branch = "feat/cpu_num" }
-# axplat-aarch64-phytium-pi = { git = "https://github.com/arceos-org/axplat_crates.git", branch = "feat/cpu_num" }
-# axplat-aarch64-qemu-virt = { git = "https://github.com/arceos-org/axplat_crates.git", branch = "feat/cpu_num" }
-# axplat-aarch64-raspi = { git = "https://github.com/arceos-org/axplat_crates.git", branch = "feat/cpu_num" }
-# axplat-loongarch64-qemu-virt = { git = "https://github.com/arceos-org/axplat_crates.git", branch = "feat/cpu_num" }
-# axplat-riscv64-qemu-virt = { git = "https://github.com/arceos-org/axplat_crates.git", branch = "feat/cpu_num" }
-# axplat-x86-pc = { git = "https://github.com/arceos-org/axplat_crates.git", branch = "feat/cpu_num" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,14 +82,14 @@ axdriver_vsock = { git = "https://github.com/arceos-org/axdriver_crates.git", ta
 axerrno = "0.1"
 axio = "0.1"
 axklib = { git = "https://github.com/arceos-hypervisor/axklib.git" } # FIXME: pin to a specific commit or tag
-axplat = { git = "https://github.com/arceos-org/axplat_crates.git", tag = "dev-v03" }
-axplat-aarch64-bsta1000b = { git = "https://github.com/arceos-org/axplat_crates.git", tag = "dev-v03" }
-axplat-aarch64-phytium-pi = { git = "https://github.com/arceos-org/axplat_crates.git", tag = "dev-v03" }
-axplat-aarch64-qemu-virt = { git = "https://github.com/arceos-org/axplat_crates.git", tag = "dev-v03" }
-axplat-aarch64-raspi = { git = "https://github.com/arceos-org/axplat_crates.git", tag = "dev-v03" }
-axplat-loongarch64-qemu-virt = { git = "https://github.com/arceos-org/axplat_crates.git", tag = "dev-v03" }
-axplat-riscv64-qemu-virt = { git = "https://github.com/arceos-org/axplat_crates.git", tag = "dev-v03" }
-axplat-x86-pc = { git = "https://github.com/arceos-org/axplat_crates.git", tag = "dev-v03" }
+axplat = { git = "https://github.com/arceos-org/axplat_crates.git", branch = "feat/cpu_num" }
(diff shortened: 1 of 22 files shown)


AsakuraMizu:
Question: What is the recommended way to configure `max-cpu-num`? Should it be overridden in the platform configuration, or modified in `defconfig.toml` (or a similar method)? Will the default configuration of `8` cause unnecessary overhead for single-core builds?

modules/axhal/src/lib.rs
```diff
@@ -124,3 +126,76 @@ pub fn init_early(cpu_id: usize, arg: usize) {
...
+/// Gets the number of CPUs running in the system.
+///
+/// When SMP is disabled, this function always returns 1.
+///
+/// When SMP is enabled, It's the smaller one between the platform-declared CPU
+/// number [`axplat::power::cpu_num`] and the configured maximum CPU number
+/// `axconfig::plat::MAX_CPU_NUM`.
+///
+/// This value is determined during the BSP initialization phase.
+pub fn cpu_num() -> usize {
```
equation314, line 144: Use `Once` or `lazy_static` to avoid manually calling the initialization function?

scripts/make/config.mk
```diff
@@ -17,7 +17,9 @@ ifneq ($(SMP),)
 else
   SMP := $(shell axconfig-gen $(PLAT_CONFIG) -r plat.cpu-num 2>/dev/null)
```
equation314: Change to `max-cpu-num`?

api/arceos_api/Cargo.toml
```diff
@@ -42,6 +42,7 @@ axipi = { workspace = true, optional = true }
 axlog.workspace = true
 axmm = { workspace = true, optional = true }
 axnet = { workspace = true, optional = true }
+axplat = { workspace = true }
```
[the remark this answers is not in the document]
equation314: Remove this?

Cargo.toml
```diff
@@ -108,3 +108,13 @@ page_table_multiarch = "0.5"
 percpu = "0.2"
 scope-local = "0.1"
 spin = "0.10"
+
+# [patch."https://github.com/arceos-org/axplat_crates.git"]
```
Azure-stars: Do they need to be removed?

modules/axhal/src/dummy.rs
```diff
@@ -110,6 +110,11 @@ impl PowerIf for DummyPower {
     fn system_off() -> ! {
         unimplemented!()
     }
+
+    #[cfg(feature = "smp")]
```
Azure-stars: Remove it?

modules/axhal/src/lib.rs
```diff
@@ -124,3 +126,76 @@ pub fn init_early(cpu_id: usize, arg: usize) {
...
+        let max_cpu_num = axconfig::MAX_CPU_NUM;
+        let cpu_num = plat_cpu_num.min(max_cpu_num);
+
+        info!(
+            "{} CPUs detected by platform, max supported is {}. {} will be used.",
+            plat_cpu_num, max_cpu_num, cpu_num
+        );
+
+        if plat_cpu_num > max_cpu_num {
+            warn!(
```
Azure-stars, line 175: Why this log uses level-warn?

modules/axhal/src/lib.rs
```diff
@@ -124,3 +126,76 @@ pub fn init_early(cpu_id: usize, arg: usize) {
...
+    }
+}
+
+/// Initializes the CPU number information for secondary CPUs.
+///
+/// Used to ensure the correct value of `CPU_NUM` is visible to secondary CPUs.
+#[cfg(feature = "smp")]
+pub fn init_cpu_num_secondary() {
+    // Ensure the correct value of CPU_NUM is visible to secondary CPUs.
+    let _ = CPU_NUM.load(Ordering::Acquire);
```
Azure-stars: Can we use `core::sync::atomic::fence(Ordering::SeqCst)` when setting the value of it and avoid calling this function?

modules/axtask/src/api.rs
```diff
@@ -87,21 +83,36 @@ pub fn current() -> CurrentTask {
...
+    // Initialize the run queue.
     crate::run_queue::init();
 
     info!("  use {} scheduler.", Scheduler::scheduler_name());
 }
 
-pub(crate) fn active_cpu_num() -> usize {
-    CPU_NUM.load(core::sync::atomic::Ordering::Relaxed)
+/// The full CPU mask of the system.
+static CPU_MASK_FULL: lazyinit::LazyInit<AxCpuMask> = lazyinit::LazyInit::new();
```
Azure-stars, line 98: Can we use `lazy_static`?

configs/dummy.toml
Azure-stars, line 1: Does plat.cpu-num need to be rename as `max-cpu-num`?

And I think we need to add appropriate comments to distinguish between max-cpu-num and plat.max-cpu-num.

aarkegz:
@equation314 @Azure-stars
I think the key point is that statically-configured and dynamically-probing platforms need different cpu number configs. Static platforms need `CPU_NUM`, which happens to be a perfect value for `MAX_CPU_NUM`. Dynamic platforms need `MAX_CPU_NUM`, and does not need `CPU_NUM` at all. I agree that there should be only one config item, but we need find a better name for it.

modules/axhal/src/dummy.rs
aarkegz: Removed

Cargo.toml
aarkegz: Removed

api/arceos_api/Cargo.toml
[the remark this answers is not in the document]
aarkegz: Removed

modules/axhal/src/lib.rs
aarkegz: `lazy_static` may lead to slightly poorer read performance?

modules/axhal/src/lib.rs
aarkegz: That make sense.

aarkegz: Maybe I'm wrong but is a single `fence` call enough to ensure visibility on another thread?

modules/axtask/src/api.rs
AsakuraMizu: > Can we use `lazy_static`?

lazy_static is almost deprecated. Should prevent usage in new code.

scripts/make/config.mk
aarkegz: Done.

aarkegz:
Closed as completed in #317 .

PR #317, opened 2025-12-18
Introduce max-cpu-num config item, remove all direct usages of axconfig::plat::CPU_NUM, use axplat to get cpu num.

aarkegz:
Replace #314, targeting main instead.

### DIFF
--- a/modules/axhal/src/lib.rs
+++ b/modules/axhal/src/lib.rs
@@ -107,6 +107,8 @@ pub use axplat::init::init_later;
 
 #[cfg(feature = "smp")]
 pub use axplat::init::{init_early_secondary, init_later_secondary};
+#[cfg(feature = "smp")]
+use core::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 
 /// Initializes CPU-local data structures for the primary core.
 ///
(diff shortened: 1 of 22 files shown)


modules/axtask/src/run_queue.rs
```diff
@@ -105,7 +106,7 @@ fn select_run_queue_index(cpumask: AxCpuMask) -> usize {
 
     // Round-robin selection of the run queue index.
     loop {
-        let index = RUN_QUEUE_INDEX.fetch_add(1, Ordering::SeqCst) % axconfig::plat::CPU_NUM;
```
equation314, line 108: Cache `cpu_num` outside the `loop`.

aarkegz: cached.

equation314 (approved):

modules/axhal/src/lib.rs
```diff
@@ -34,6 +34,10 @@
 #[macro_use]
 extern crate log;
 
+#[allow(unused_imports)]
+#[macro_use]
+extern crate axlog;
+
```
AsakuraMizu, line 40: Is this useful?

aarkegz: Yes, to log smp properly.

AsakuraMizu: Ok I see you called `ax_println` in `init_cpu_num`. A bit strange...

PR #351, opened 2026-03-06
fix: axplat-dyn build on default target

ZR233:


AsakuraMizu:
The suggestions in https://github.com/arceos-org/arceos/pull/351#pullrequestreview-3901579170 are good.

PR #360, opened 2026-03-18
feat(ci): add GitHub Actions workflow to notify parent repository on …

ZCShou:
…updates


equation314:
Could it be changed so that the parent repository periodically checks for updates in the sub-repositories? The current approach requires adding workflows to all sub-repositories, resulting in poor scalability.

AsakuraMizu:
> Could it be changed so that the parent repository periodically checks for updates in the sub-repositories? The current approach requires adding workflows to all sub-repositories, resulting in poor scalability.

First, I've suggested @ZCShou to use the `workflow_call` approach to simplify the configuration yaml in each sub repo, which may make it better. And you've inspired me that if we can use [webhooks](https://docs.github.com/en/webhooks/using-webhooks).

